PR #27, opened 2020-01-15
Convert to vue

### LINKED ISSUE #26: Convert to vue

### LINKED ISSUE #28: Fix issue with disallowed characters in the css selectors
There are certain characters not allowed in the css selectors that we should check for. This includes "." which I fudged in #27 by removing the "." from all ids in multinet.js

JackWilb:
Closes #26.

TODO:
- [x] Draw all labels
- [x] Add highlighting back (css selectors are broken)
- [x] There's a bug in the hover highlight sometime it doesn't render both bars (temp solution is to remove "." from ids, opened #28) (closes #28 with help from Roni on the selectors)
- [x] Add the sorting svg elements back
- [x] Add the sorting algorithms back
- [x] Clean all linter errors
- [ ] Fix the re-render problems
- [x] Make some of the variables props so we can update the MVC (opened #30)


JackWilb:
Hey @waxlamp, this is ready for review. There are a couple of stubborn linter errors that I can't get rid of in the model.ts file. If you could provide some insight into how to fix them, that'd be great!

src/components/AdjMatrix/AdjMatrix.vue
```diff
@@ -0,0 +1,216 @@
...
+        graphStructure,
+      };
+    },
+  },
+
+  watch: {
+    properties() {
+      //   this.updateVis();
+    },
+  },
```
waxlamp: If you don't need to react to changes in the "properties" (which I mention because the watcher's body is commented out) then I'd get rid of this block of code.

src/components/AdjMatrix/AdjMatrix.vue
```diff
@@ -0,0 +1,216 @@
...
+};
+</script>
+
+<template>
+  <div>
+    <svg ref="svg" width="800" height="900" />
+  </div>
+</template>
+
+<style>
```
waxlamp:
```suggestion
<style scoped>
```

You should probably use scoped styles for this; if this component ever gets embedded somewhere else, this could cause headache-inducing style bugs.

src/components/AdjMatrix/controller.ts
```diff
@@ -0,0 +1,92 @@
+/* The Controller controls the model and tell the model how to manipulate the data */
```
waxlamp: It does seem like this functionality could be distributed into the component at the Vue level but it's not critical (especially for this PR).

src/lib/multinet.js
```diff
@@ -0,0 +1,88 @@
+/* Multinet data importer */
```
waxlamp: As soon as possible, we should update these functions to use multinetjs rather than raw API calls.

JackWilb:
> There are couple of empty files (`cleaned_model.js` and `cleaned_model.ts`, I think) that ought to be removed before merging.

I removed these. Does that not show for you on the branch?

waxlamp:
> > There are couple of empty files (`cleaned_model.js` and `cleaned_model.ts`, I think) that ought to be removed before merging.
> 
> I removed these. Does that not show for you on the branch?

Weird. The "files changed" view shows both files with 0 bytes and say "no change" rather than "deleted". When I check the branch out, I indeed don't see those files. I wonder what's going on, but it doesn't matter for the PR itself--carry on :)

src/components/AdjMatrix/AdjMatrix.vue
JackWilb: I left this here because this is where the control panel properties will trigger the updates

src/components/AdjMatrix/controller.ts
JackWilb: I agree. We can probably look to migrate more and more of this code to be in the .vue files over time but just to get it working for now, this is okay

src/lib/multinet.js
JackWilb: Agreed. As soon as [#188](https://github.com/multinet-app/multinet/issues/188) is merged we can go for it

src/components/AdjMatrix/AdjMatrix.vue
JackWilb: This will require moving some of these rules up to controls.vue, etc. I think it's important we get this right now so I'll look into this change

waxlamp: Might be better to do it in a followup PR, then. If everything works right now, then getting this merged and making incremental improvements is probably the way to go.

JackWilb: Roger that. Opened #29

PR #32, opened 2020-01-27
Fix build errors, type errors, and some html updates

JackWilb:


src/lib/provenance.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+import * as Provenance from 'provenance-lib-core/lib/src/provenance-core/Provenance';
+
+function _nodeLink(provenance: Provenance.Provenance<{
+  order: string; nodePos: {}; // map of node positions,
+    userSelectedNeighbors: never[]; // map of nodes that have neighbors selected (so they can be non-muted)
+    userSelectedEdges: never[];
+    selected: never[]; // set of nodes that have been 'soft selected'
+    hardSelected: never[]; // set of nodes that have been 'hard selected'
+    search: never[]; // field to store the id of a searched node;
```
waxlamp: All these `never`s seems a little weird. Are these values just not used in the rest of the codebase?

JackWilb: That was automatically inferred and I couldn't get it to work any other way. The app functions without this code so I just removed it for now

PR #36, opened 2020-01-28
Iterative improvement to typing and removal of some dead code

### LINKED ISSUE #33: Define better type hints by substituting the any types
There are lots of types that could be more precisely defined. Doing this better typing will help with supporting the software in the future

JackWilb:
First of a few iterative improvements for #33 
Must be closed after #35 and #34 (They're already merged)

Also fixes some loose ends found along the way

### DIFF
--- a/src/components/AdjMatrix/controller.ts
+++ b/src/components/AdjMatrix/controller.ts
@@ -2,12 +2,6 @@
 import * as d3 from 'd3';
 
 export class Controller {
-  public columnSelectedNodes: {} = {};
-  public highlightedNodes: {} = {};
-  public attributeProportion: number;
-  public sortKey: string;
-  public nodeAttributes: any;
-
   private view: any;
   private model: any;
   private clickedCells: any;
@@ -15,6 +9,11 @@ export class Controller {
   private hoverCol: any;
   private datumID: any;
   private visDimensions: any;
+  private columnSelectedNodes: {} = {};
+  private sortKey: string;
+  private nodeAttributes: any;
+  private highlightedNodes: {} = {};
+  private attributeProportion: number;
 
   constructor(view: any, model: any, visDimensions: any) {
     this.visDimensions = visDimensions;
@@ -37,7 +36,8 @@ export class Controller {
 
     this.model.reload();
 
-    this.changeOrder('shortName', false);
+    this.sortKey = 'shortName';
+    this.changeOrder(this.sortKey, false);
   }
 
   public clear() {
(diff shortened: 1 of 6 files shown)


src/components/AdjMatrix/controller.ts
```diff
@@ -2,19 +2,18 @@
...
   private model: any;
-  private sortKey: any;
   private clickedCells: any;
   private hoverRow: any;
   private hoverCol: any;
   private datumID: any;
   private visDimensions: any;
+  private columnSelectedNodes: {} = {};
+  private sortKey: string;
+  private nodeAttributes: any;
```
waxlamp: This sounds like it should be an array maybe?

src/components/AdjMatrix/controller.ts
```diff
@@ -2,19 +2,18 @@
...
+  private highlightedNodes: {} = {};
```
waxlamp: This seems like it might have a more specific type; perhaps a mapping from names to some particular data structure to describe the highlighted nodes?

src/components/AdjMatrix/model.ts
```diff
@@ -4,23 +4,28 @@ import * as ProvenanceLibrary from 'provenance-lib-core/lib/src/provenance-core/
 import 'science';
 import 'reorder.js';
 
-declare var reorder: any;
+declare const reorder: any;
 
 export class Model {
-  public graphStructure: any;
+  public graphStructure: {nodes: object[], links: object[]};
```
waxlamp: Similar comment here: seems like instead of `object[]` there might be a more specific array type.

src/components/AdjMatrix/controller.ts
JackWilb: Actually looks like it's not used, removed it.

src/components/AdjMatrix/controller.ts
JackWilb: Updated so that the keys point to any[]

JackWilb: Actually, to `{ [key: string]: Array<{ [key: string]: any}> }`

src/components/AdjMatrix/model.ts
JackWilb: Yeah, I'll address this in a follow up with the types.ts file

JackWilb:
> There are still a few instances of `any` lying around; if these are intentional, it's a good idea to add a comment explaining why.
> 
> I'm approving this PR because the changes are looking good, although I don't understand the codebase well enough to really understand what's going on :-O
> 
> And one suggestion: especially for some of the more verbose types, but generally for all semantically rich types, I would suggest creating type definitions in a central place (usually `types.ts` in the top level) and then referring to these by name throughout the codebase. That simple refactoring does _wonders_ for code legibility.

I agree with you on a global types file, but I just wanted to add a couple more types here and remove some dead code. My game plan was to iterate a few smaller changes. I'll leave the issue open still and open a new one for the types.ts file :)

PR #40, opened 2020-02-21
Implement the attributes view into the adjacency matrix

### LINKED ISSUE #16: Implement the attribute visuals next to the matrix
Alex and I had talked about maybe having selections, adding, sorting, etc. similar to [taggle](https://taggle.caleydoapp.org/#aids).

JackWilb:
Closes #16

### DIFF
--- a/src/components/AdjMatrix/AdjMatrix.vue
+++ b/src/components/AdjMatrix/AdjMatrix.vue
@@ -28,10 +28,6 @@ export default Vue.extend({
       type: Array,
       default: [],
     },
-    variableList: {
-      type: Array,
-      default: [],
-    },
   },
 
   data(): {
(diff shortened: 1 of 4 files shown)


JackWilb:
@waxlamp, this is sufficiently far to be tested locally. Let me know what you think of the styling and if you have any recommendations for dealing with the attribute label length

JackWilb:
@waxlamp This is now rendering numeric and categorical data and has tooltips, etc. Let me know how you like it so far. There is a bit of work still to be done regarding enter, exit, merge with the D3; I've kind of just hacked that together for now. I wanted to get some feedback before I focus on that though.

waxlamp:
This is looking great. When I load up the les mis data, I see that the attributes view overplots a bit on the last column from the matrix view (or maybe not the very last column; hard to tell which columns are obscured just by looking).

There is probably a bit of design help this could use, and I can loop in @jtomeck if you like to take a look.

JackWilb:
@waxlamp I just fixed up that slight issue with the extra row. It was a problem with the margin calculation. I'm happy to loop Jared in but maybe we should do that as a separate PR since I'm sure that will give yield substantial changes

PR #41, opened 2020-02-21
Add yarn caching to speed up CI builds

### LINKED ISSUE #38: Implement yarn and pipenv caching for the CI
Using [this](https://github.com/actions/cache) action we should start caching to speed up our builds. The yarn specific instructions are [here](https://github.com/actions/cache/blob/master/examples.md#node---yarn) and the pipenv instructions (not fully implemented yet) are [here](https://github.com/actions/cache/issues/152).

JackWilb:
Closes #38

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,6 +18,22 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: '10.x'
-    - run: yarn install
+
+    - name: Get yarn cache directory path
+      id: yarn-cache-dir-path
+      run: echo "::set-output name=dir::$(yarn cache dir)"
+
+    - uses: actions/cache@v1
+      id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+      with:
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+        key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-yarn-
+
+    - name: Install yarn packages
+      if: steps.yarn-cache.outputs.cache-hit != 'true'
+      run: yarn install
+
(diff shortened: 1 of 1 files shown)


JackWilb:
This should be ready to go @waxlamp. You can verify that it worked on the 3rd to last and second to last CI runs. It used the cache reducing the install time of node_modules from 25 seconds to 5. In the last run I updated the name of the cache (from the temporary debugging name that I was using) so it might won't have restored the cache but should work into the future.

.github/workflows/main.yml
```diff
@@ -18,6 +18,15 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: '10.x'
-    - run: yarn install
+
+    - uses: actions/cache@v1
+      id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
```
waxlamp:
```suggestion
      id: yarn-cache
```

.github/workflows/main.yml
```diff
@@ -18,6 +18,15 @@ jobs:
...
+      with:
+        path: /home/runner/work/view-adjmatrix/view-adjmatrix/node_modules
+        key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}-cache
```
waxlamp: How did you know to use this string as the key?

And also, what is that `hashFiles()` call doing? I would have thought that, since we know where the `yarn.lock` file is, we could just refer to it directly?

JackWilb: This comes exactly from the example in the actions repo. The game plan was to create a hash from the lock file that changes as the file updates. The key is the unique identifier for the cache so we want it to be tied to the lock file.

JackWilb: The exact code for the hash function is [here](https://github.com/actions/runner/blob/master/src/Runner.Worker/ExpressionFunctions/HashFiles.cs)

PR #62, opened 2020-05-22
Update naming conventions to match project naming conventions

### LINKED ISSUE #58: Change "variable" to "attribute" in UI
The proper term in a Multinet context is "attribute" rather than "variable".

### LINKED ISSUE #61: Use the word network instead of graph to align better with our nomenclature
Inspired by #58

JackWilb:
Closes #58
Closes #61

This will currently break the deployment since the query string parameter is graph, not network. 

Opened multinet-app/multinet-client#90 to switch from graph in the querystring to network

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -1,14 +1,14 @@
 // Get the url querystring variables
 export function getUrlVars(): { [key: string]: string} {
   const workspace: string = '';
-  const network: string = '';
-  const vars: { [key: string]: string} = {workspace, network};
+  const graph: string = '';
+  const vars: { [key: string]: string} = {workspace, graph};
   window.location.href.replace(/[?&]+([^=&]+)=([^&*#]*)/gi,
     (m: any, key: string, value: any) => {
       vars[key] = value;
       return value;
     },
   );
-  vars.networkName = vars.network;
+  vars.networkName = vars.graph;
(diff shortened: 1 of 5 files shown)


src/lib/utils.ts
```diff
@@ -1,13 +1,14 @@
...
-  const vars: { [key: string]: string} = {workspace, graph};
+  const network: string = '';
+  const vars: { [key: string]: string} = {workspace, network};
   window.location.href.replace(/[?&]+([^=&]+)=([^&*#]*)/gi,
     (m: any, key: string, value: any) => {
       vars[key] = value;
       return value;
     },
   );
+  vars.networkName = vars.network;
```
JackWilb, line 12: This doesn't seem ideal but is necessary because the network variable in the main application is the actual network object. We could mitigate this by requiring that the parameter in the query string is `networkName=` and not `network=`, but let me know what you think @waxlamp

waxlamp: The query string should say `network`, not `networkName`, because it's sort of part of the external interface to this application.

What's wrong with doing it this way?

JackWilb: There's no real problem, it just feels a bit like shadow naming the variable. This problem is removed in 4a0a6ce where we just use the parameter graph

Issue #59, opened 2020-05-22
Show node and link data on hover


waxlamp:
Perhaps using a popover or tooltip.

JackWilb:
We currently have a tooltip that does this

PR #63, opened 2020-05-27
Lots of minor fixes

### LINKED ISSUE #33: Define better type hints by substituting the any types

### LINKED ISSUE #37: There is an additional row of cells for both the rows and columns
Must be an off by one error somewhere, they're not clickable.

JackWilb:
Fixes a lot of minor issues:
- Tooltip positioning
- D3 does not plot outside the SVG (I think this was causing #37 so this closes #37)
- Adds better typing (closes #33)
- Refactors the hovering logic

Partial fix for #59.

These fixes simplify some of the hovering logic which will be important for debugging the slow highlighting

### DIFF
--- a/src/components/AdjMatrix/AdjMatrix.vue
+++ b/src/components/AdjMatrix/AdjMatrix.vue
@@ -85,15 +85,15 @@ export default Vue.extend({
     // Size the svgs
     this.matrix = d3
       .select(this.$refs.matrix)
-      .attr('width', this.visDimensions.width * 0.75 - 30)
+      .attr('width', this.visDimensions.width * 0.75)
       .attr('height', this.visDimensions.height)
-      .attr('viewBox', `0 0 ${this.visDimensions.width * 0.75 - 30} ${this.visDimensions.height}`);
+      .attr('viewBox', `0 0 ${this.visDimensions.width * 0.75} ${this.visDimensions.height}`);
 
     this.attributes = d3
       .select(this.$refs.attributes)
-      .attr('width', this.visDimensions.width * 0.25 - 30)
+      .attr('width', this.visDimensions.width * 0.25)
       .attr('height', this.visDimensions.height)
-      .attr('viewBox', `0 0 ${this.visDimensions.width * 0.25 - 30} ${this.visDimensions.height}`);
+      .attr('viewBox', `0 0 ${this.visDimensions.width * 0.25} ${this.visDimensions.height}`);
 
     // Define the View
     this.view = new View(this.network, this.visDimensions, this.visualizedAttributes);
--- a/src/components/AdjMatrix/AdjMatrix.vue
+++ b/src/components/AdjMatrix/AdjMatrix.vue
@@ -91,9 +91,9 @@ export default Vue.extend({
 
     this.attributes = d3
       .select(this.$refs.attributes)
-      .attr('width', this.visDimensions.width * 0.25)
+      .attr('width', this.visDimensions.width * 0.25 - 15) // -15 for the scroll bar
       .attr('height', this.visDimensions.height)
-      .attr('viewBox', `0 0 ${this.visDimensions.width * 0.25} ${this.visDimensions.height}`);
+      .attr('viewBox', `0 0 ${this.visDimensions.width * 0.25 - 15} ${this.visDimensions.height}`);
 
     // Define the View
     this.view = new View(this.network, this.visDimensions, this.visualizedAttributes);
--- a/src/components/AdjMatrix/AdjMatrixMethods.ts
+++ b/src/components/AdjMatrix/AdjMatrixMethods.ts
@@ -201,18 +201,18 @@ export class View {
       .merge(path)
       .attr('class', `sortIcon attr attrSortIcon`)
       .attr('cursor', 'pointer')
-      .attr('d', (d: any) => {
+      .attr('d', (d: string) => {
         const type = this.isQuantitative(d) ? 'quant' : 'categorical';
         return this.icons[type].d;
       })
-      .attr('transform', (d: any, i: number) => `scale(0.1)translate(${(colWidth + this.colMargin) * i * 10 - 200}, -1100)`)
-      .style('fill', (d: any) => '#8B8B8B')
-      .on('click', (d: any) => this.sort(d));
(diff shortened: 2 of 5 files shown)


src/components/AdjMatrix/AdjMatrix.vue
```diff
@@ -85,15 +85,15 @@ export default Vue.extend({
     // Size the svgs
     this.matrix = d3
       .select(this.$refs.matrix)
-      .attr('width', this.visDimensions.width * 0.75 - 30)
+      .attr('width', this.visDimensions.width * 0.75)
```
waxlamp, line 88: I think it would make sense to refactor all of these values into computed props with a descriptive name (like `matrixWidth` or something like that).

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -41,13 +32,10 @@ export class View {
   private colMargin: number = 5;
   private visDimensions: Dimensions;
   private provenance: any;
-  private idMap: { [key: string]: number};
-  private hoverRow: {} = {};
-  private hoverCol: {} = {};
+  private idMap: { [key: string]: number };
   private isMultiEdge: any;
```
waxlamp, line 40: `isMultiEdge` sounds like a `boolean`.

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -82,20 +71,13 @@ export class View {
...
-   * @param  data [description]
-   * @return      [description]
-   */
-  public loadData(): void {
     // Kick off the rendering
     this.initializeEdges();
     this.initializeAttributes();
   }
 
+
```
waxlamp, line 84:
```suggestion

```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -942,15 +896,15 @@ export class View {
...
+  private nodeClick(node: Node, search: boolean = false): void {
   const previousState = this.getApplicationState();
   let clicked = previousState.clicked;
   const wasSelected = this.isSelected(node);
 
   if (wasSelected) {
     clicked = clicked.filter((s: any) => s !== node.id);
   } else {
-    clicked.push(node.id);
+    clicked.push(node.id as never);
```
waxlamp, line 760: What does this mean? (I haven't run across the `never` type too much.)

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -1232,66 +1186,35 @@ export class View {
...
-            }
-          });
+    // Find max value of z
+    this.matrix.forEach((row: Cell[]) => {
+      row.forEach((cell: Cell) => {
+        if (cell.z > this.maxNumConnections) {
+          this.maxNumConnections = cell.z;
+        }
       });
+    });
```
waxlamp, line 1020: There is probably a more functional way to extract this max value. It may not be worthwhile to convert it over to that style though. Let me know if you want to discuss it.

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -625,142 +585,36 @@ export class View {
...
+      for (const neighborNode of this.columnSelectedNodes[node]) {
+        cssSelector += `
+        [id="attrRow${neighborNode}"],[id="topoRow${neighborNode}"],[id="nodeLabelRow${neighborNode},"]
+        `;
       }
     }
-    // remove last comma
+
+    // Remove last comma
     cssSelector = cssSelector.substring(0, cssSelector.length - 1);
```
waxlamp: This feels pretty brittle. You may want to consider generating an array of strings to hold the individual selectors, then calling `join()` on it at the end.

src/components/AdjMatrix/AdjMatrix.vue
```diff
@@ -91,9 +91,9 @@ export default Vue.extend({
 
     this.attributes = d3
       .select(this.$refs.attributes)
-      .attr('width', this.visDimensions.width * 0.25)
+      .attr('width', this.visDimensions.width * 0.25 - 15) // -15 for the scroll bar
       .attr('height', this.visDimensions.height)
-      .attr('viewBox', `0 0 ${this.visDimensions.width * 0.25} ${this.visDimensions.height}`);
+      .attr('viewBox', `0 0 ${this.visDimensions.width * 0.25 - 15} ${this.visDimensions.height}`);
```
waxlamp: I would recommend keeping the `15` in a `const` variable (or even a computed prop, but that might be too extreme).

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: This means that the clicked array was never defined. I just added `clicked: string[] = previousState.clicked;` and now I can pass it without a type conversion

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: Fixed on my new branch at 006792d

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: My preference is 2 spaces between function blocks. Fixed on my new branch at 006792d

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: Fixed on my new branch at 006792d

src/components/AdjMatrix/AdjMatrix.vue
JackWilb: I refactored this on my new branch at 8774837

src/components/AdjMatrix/AdjMatrix.vue
JackWilb: I didn't store this in a const because with the refactoring in 8774837, it's only used in one place

src/components/AdjMatrix/AdjMatrixMethods.ts
waxlamp: No, it's more just a style thing. I find that functional thinking leads to better code, both in terms of readability and reasoning about it if something goes wrong. Since this works just fine, there's no rush to change it, more just something to think about.

PR #64, opened 2020-06-01
Fix hover rendering slowness

JackWilb:
Depends on #63 

There are a couple of ideas we had to reduce the hover rendering slowness:

Tried and in the PR:
- Reducing the number of DOM elements (> 10,000 from the cell rendering logic on the miserables graph)

Proposed solutions:
- [ ] Changing the selector
- [ ] Only rendering cells that have a link
These were not tested since reducing the number of elements seems to have fixed the problem.

### DIFF
--- a/src/components/AdjMatrix/AdjMatrix.vue
+++ b/src/components/AdjMatrix/AdjMatrix.vue
@@ -62,19 +62,19 @@ export default Vue.extend({
       };
     },
 
-    matrixWidth() {
+    matrixWidth(): number {
       return this.visDimensions.width * 0.75;
     },
 
-    matrixHeight() {
+    matrixHeight(): number {
       return this.visDimensions.height;
     },
 
-    attributesWidth() {
+    attributesWidth(): number {
       return this.visDimensions.width * 0.25 - 15; // 15 for the scrollbar
     },
 
-    attributesHeight() {
+    attributesHeight(): number {
       return this.matrixHeight;
     },
   },
--- a/src/components/AdjMatrix/AdjMatrix.vue
+++ b/src/components/AdjMatrix/AdjMatrix.vue
@@ -112,7 +112,12 @@ export default Vue.extend({
       .attr('viewBox', `0 0 ${this.attributesWidth} ${this.attributesHeight}`);
 
     // Define the View
-    this.view = new View(this.network, this.visDimensions, this.visualizedAttributes);
+    this.view = new View(
+      this.network,
+      this.visualizedAttributes, 
+      this.matrixWidth, 
+      this.matrixHeight
+    );
   },
 
   methods: {
--- a/src/components/AdjMatrix/AdjMatrix.vue
+++ b/src/components/AdjMatrix/AdjMatrix.vue
@@ -61,6 +61,22 @@ export default Vue.extend({
         visualizedAttributes,
       };
     },
+
+    matrixWidth() {
+      return this.visDimensions.width * 0.75;
+    },
+
+    matrixHeight() {
+      return this.visDimensions.height;
+    },
+
+    attributesWidth() {
+      return this.visDimensions.width * 0.25 - 15; // 15 for the scrollbar
+    },
+
+    attributesHeight() {
+      return this.matrixHeight;
+    },
   },
 
   watch: {
@@ -85,15 +101,15 @@ export default Vue.extend({
     // Size the svgs
     this.matrix = d3
       .select(this.$refs.matrix)
-      .attr('width', this.visDimensions.width * 0.75)
-      .attr('height', this.visDimensions.height)
-      .attr('viewBox', `0 0 ${this.visDimensions.width * 0.75} ${this.visDimensions.height}`);
+      .attr('width', this.matrixWidth)
+      .attr('height', this.matrixHeight)
+      .attr('viewBox', `0 0 ${this.matrixWidth} ${this.matrixHeight}`);
 
     this.attributes = d3
       .select(this.$refs.attributes)
-      .attr('width', this.visDimensions.width * 0.25 - 15) // -15 for the scroll bar
-      .attr('height', this.visDimensions.height)
-      .attr('viewBox', `0 0 ${this.visDimensions.width * 0.25 - 15} ${this.visDimensions.height}`);
+      .attr('width', this.attributesWidth) // -15 for the scroll bar
+      .attr('height', this.attributesHeight)
+      .attr('viewBox', `0 0 ${this.attributesWidth} ${this.attributesHeight}`);
 
     // Define the View
     this.view = new View(this.network, this.visDimensions, this.visualizedAttributes);
--- a/src/components/AdjMatrix/AdjMatrix.vue
+++ b/src/components/AdjMatrix/AdjMatrix.vue
@@ -107,7 +107,7 @@ export default Vue.extend({
 
     this.attributes = d3
       .select(this.$refs.attributes)
-      .attr('width', this.attributesWidth) // -15 for the scroll bar
+      .attr('width', this.attributesWidth)
       .attr('height', this.attributesHeight)
       .attr('viewBox', `0 0 ${this.attributesWidth} ${this.attributesHeight}`);
 
--- a/src/components/AdjMatrix/AdjMatrix.vue
+++ b/src/components/AdjMatrix/AdjMatrix.vue
@@ -167,18 +167,13 @@ svg >>> .hovered {
   fill-opacity: 1 !important;
 }
 
-svg >>> .clickedCell {
-  stroke-width: 2.5;
-  stroke: #ee0000;
-}
-
 svg >>> .clicked {
   font-weight: 800;
   fill: #f8cf91 !important;
   fill-opacity: 1 !important;
 }
 
-svg >>> .clickedCell {
+svg >>> .cell.clicked {
   stroke: red;
   stroke-width: 3;
 }
--- a/src/components/AdjMatrix/AdjMatrixMethods.ts
+++ b/src/components/AdjMatrix/AdjMatrixMethods.ts
@@ -290,17 +290,17 @@ export class View {
       .domain([0, this.maxNumConnections])
       .range(['#feebe2', '#690000']); // TODO: colors here are arbitrary, change later
 
-    const cells = this.edgeRows.selectAll('.cell')
+    this.edgeRows.selectAll('.cell')
       .data((d: Node, i: number) => this.matrix[i])
       .enter()
       .append('rect')
+      .attr('class', 'cell')
+      .attr('id', (d: Cell) => d.cellName)
       .attr('x', (d: Cell) => this.orderingScale(d.x))
       .attr('width', this.orderingScale.bandwidth())
       .attr('height', this.orderingScale.bandwidth())
       .style('fill', (d: Cell) => cellColorScale(d.z))
-      .style('fill-opacity', (d: Cell) => d.z);
-
-    cells
+      .style('fill-opacity', (d: Cell) => d.z)
       .on('mouseover', (d: Cell, i: number, nodes: any) => {
         this.showToolTip(d, i, nodes);
         this.hoverEdge(d);
@@ -309,11 +309,7 @@ export class View {
         this.hideToolTip();
         this.unHoverEdge(d);
       })
-      .on('click', (d: Cell, i: number, nodes: any) => {
-        const interaction = select(nodes[i]).attr('class');
-        const action = this.changeInteractionWrapper(interaction, d);
-        this.provenance.applyAction(action);
-      })
+      .on('click', (d: Cell) => this.cellClick(d))
       .attr('cursor', 'pointer');
 
     this.appendEdgeLabels();
--- a/src/components/AdjMatrix/AdjMatrixMethods.ts
+++ b/src/components/AdjMatrix/AdjMatrixMethods.ts
@@ -34,7 +34,6 @@ export class View {
   private columnHeaders: any;
   private attributeScales: { [key: string]: any } = {};
   private colMargin: number = 5;
-  private visDimensions: Dimensions;
   private provenance: any;
   private idMap: { [key: string]: number };
   private isMultiEdge: boolean;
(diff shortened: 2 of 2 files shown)


src/components/AdjMatrix/AdjMatrix.vue
```diff
@@ -85,15 +101,15 @@ export default Vue.extend({
...
+      .attr('width', this.matrixWidth)
+      .attr('height', this.matrixHeight)
+      .attr('viewBox', `0 0 ${this.matrixWidth} ${this.matrixHeight}`);
 
     this.attributes = d3
       .select(this.$refs.attributes)
-      .attr('width', this.visDimensions.width * 0.25 - 15) // -15 for the scroll bar
-      .attr('height', this.visDimensions.height)
-      .attr('viewBox', `0 0 ${this.visDimensions.width * 0.25 - 15} ${this.visDimensions.height}`);
+      .attr('width', this.attributesWidth) // -15 for the scroll bar
```
waxlamp, line 110:
```suggestion
      .attr('width', this.attributesWidth)
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -215,19 +218,21 @@ export class View {
 
   }
 
+
```
waxlamp, line 221:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -215,19 +218,21 @@ export class View {
...
   private isQuantitative(varName: string): boolean {
     const uniqueValues = [...new Set(this.network.nodes.map((node: Node) => parseFloat(node[varName])))];
     return uniqueValues.length > 5;
   }
 
+
```
waxlamp, line 227:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -215,19 +218,21 @@ export class View {
...
   /**
    * initializes the edges view, renders all SVG elements and attaches listeners
    * to elements.
    * @return None
    */
   private initializeEdges(): void {
     // Set width and height based upon the calculated layout size. Grab the smaller of the 2
-    const sideLength = Math.min(this.visDimensions.width, this.visDimensions.height);
+    const sideLength = Math.min(this.visDimensions.width * 0.75, this.visDimensions.height);
```
waxlamp, line 235: Isn't this just `matrixWidth` now?

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -324,11 +314,7 @@ export class View {
         this.hideToolTip();
         this.unHoverEdge(d);
       })
-      .on('click', (d: Cell, i: number, nodes: any) => {
-        const interaction = select(nodes[i]).attr('class');
-        const action = this.changeInteractionWrapper(interaction, d);
-        this.provenance.applyAction(action);
-      })
+      .on('click', (d: Cell) => this.selectElement(d))
```
waxlamp, line 317: I believe you can simplify this to:

```suggestion
      .on('click', this.selectElement)
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -341,6 +327,7 @@ export class View {
       .style('opacity', 0);
   }
 
+
```
waxlamp, line 330:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -408,6 +395,61 @@ export class View {
     this.unHoverNode(cell.colID);
   }
 
+
```
waxlamp, line 398:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -408,6 +395,61 @@ export class View {
...
+      } else {
+        // Get all the elements to be selected
+        elementsToSelect = [
+        `[id="attrRow${element.colID}"]`, `[id="topoRow${element.colID}"]`, `[id="topoCol${element.colID}"]`,
+        `[id="colLabel${element.colID}"]`, `[id="rowLabel${element.colID}"]`,
+
+        `[id="attrRow${element.rowID}"]`, `[id="topoRow${element.rowID}"]`, `[id="topoCol${element.rowID}"]`,
+        `[id="colLabel${element.rowID}"]`, `[id="rowLabel${element.rowID}"]`,
+
+        `[id="${element.cellName}"]`,
```
waxlamp, line 416: I have a suspicion that using attribute selectors instead of ID selectors may have a performance impact, since we're doing ~10 of them on every selection event. It would be a worthwhile experiment to do some profiling, then change all of these out for `#...` selectors to see if the performance within this function changes at all.

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -408,6 +395,61 @@ export class View {
...
+        selections.push(elementToSelect);
+      }
+    }
+
+    if (selections.length > 0) {
+      selectAll(selections.join(',')).classed('clicked', true);
+    }
+  }
+
+
```
waxlamp, line 452:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -515,6 +557,7 @@ export class View {
 
   }
 
+
```
waxlamp, line 560:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -560,6 +603,7 @@ export class View {
     };
   }
 
+
```
waxlamp, line 606:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -591,73 +635,76 @@ export class View {
...
-    const nodeIDs = this.network.nodes.map((node: { id: any; }) => node.id);
+    const nodeIDs = this.network.nodes.map((node: Node) => node.id);
 
     this.order = this.changeOrder(order, nodeIDs.includes(order));
     this.orderingScale.domain(this.order);
 
     const transitionTime = 500;
 
-    this.edgeRows
+    (selectAll('.rowContainer') as any)
```
waxlamp, line 677: Why cast to `any`?

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -748,52 +795,14 @@ export class View {
...
 
   private isSelected(node: Node): boolean {
     const currentState = this.getApplicationState();
-    const clicked = currentState.clicked;
-    return clicked.includes(node.id as never);
+    const clicked: string[] = currentState.clicked;
+    return clicked.includes(node.id);
   }
 
+
```
waxlamp, line 805:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -810,11 +819,13 @@ export class View {
       .style('opacity', .9);
   }
 
+
```
waxlamp, line 822:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -810,11 +819,13 @@ export class View {
...
   private hideToolTip(): void {
     this.tooltip.transition(25)
     .style('opacity', 0);
   }
 
+
```
waxlamp, line 828:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -863,6 +874,7 @@ export class View {
     return order;
   }
 
+
```
waxlamp, line 877:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -892,98 +904,10 @@ export class View {
...
-      provenance.addObserver('selections.neighborSelect', updateHighlights);
-      provenance.addObserver('selections.cellCol', updateCellClicks);
-      provenance.addObserver('selections.search', updateHighlights);
-      provenance.addObserver('clicked', updateHighlights);
-    }
-    setUpObservers();
     return provenance;
   }
 
+
```
waxlamp, line 910:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -1020,6 +944,7 @@ export class View {
     });
   }
 
+
```
waxlamp, line 947:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -1028,7 +953,8 @@ export class View {
     return this.provenance.graph().current.state;
   }
 
-  private changeOrder(type: string, node: boolean = false): number[] {
+
```
waxlamp, line 956:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -1054,4 +980,9 @@ export class View {
       args: [sortKey],
     };
   }
+
+
```
waxlamp, line 984:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: I prefer 2 newlines between function blocks

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: It is, but that's not passed through to the methods. I could if that would make more sense, but I believe this is the only place where we have to recalculate it

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: I don't think you can use the # selectors with our current id generation strategy (just taking the node and cell ids and making them the html attributes), because there is some issue with having either `_` or `/` in the name. Thus `#mis_nodes/child1` isn't a valid selector. So if we want to refactor to try the # selector, it's a little more involved

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: Without it, we get: 

```
Property 'transition' does not exist on type 'Selection<BaseType, unknown, HTMLElement, any>'.ts(2339)
```

Is there an easy way to update the types for this?

src/components/AdjMatrix/AdjMatrixMethods.ts
waxlamp: But you defined `matrixWidth` as a computed prop right? Then it's definitely available in the methods as `this.matrixWidth`, same as `this.visDimensions` is.

This is the kind of thing that almost certainly leads to bugs later on, when you update the matrix width calculation but forget that there's a stray manual calculation here.

src/components/AdjMatrix/AdjMatrixMethods.ts
waxlamp: I do too, if it's Python code. In JS/TS, it's not a standard thing to do, and standards (even de facto ones) matter in shared codebases.

This is of course your choice, but if you decide to add a formatter like `prettier` (which you should almost certainly do), then you can run that to format your code according to community specs, and then never worry about how to format again.

It's a bit of a change because the formatter will make choices that you don't agree with. But in my experience I've found the loss of that level of control is ultimately empowering: it means you never have to think about formatting, and it doesn't need to be a factor in code reviews either. And an unexpected benefit is that, if the formatter ever makes your code truly ugly, it is a clue that you should do some kind of refactoring in order to make the code easier to understand (and therefore, easier to format well).

src/components/AdjMatrix/AdjMatrixMethods.ts
waxlamp: If you can make this change then I'll approve the PR. Let me know if there's any trouble in doing so.

JackWilb: > But you defined `matrixWidth` as a computed prop right? Then it's definitely available in the methods as `this.matrixWidth`, same as `this.visDimensions` is.

That's passed through on initialization. I'm happy to pass through the computed props so we can get this merged.

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: Okay. I'll plan to add a formatter at some point and this should fix all of these. For now I'll keep it as is

src/components/AdjMatrix/AdjMatrixMethods.ts
waxlamp: We can easily change how the IDs are set. If the current ID values are not valid IDs, then we should fix it in any case.

I can file an issue to track this, but I think it would be worth a quick experiment to compare the performance.

waxlamp: See #68.

src/components/AdjMatrix/AdjMatrixMethods.ts
waxlamp: I think the d3 type modules are pretty bad. We can look into this later.

src/components/AdjMatrix/AdjMatrixMethods.ts
waxlamp: I may have misunderstood what is going on in the code. Using a proxy class for the view methods seems non-ideal, precisely because of the inability to use computed props, etc., from the main component, but this is fine for now.

PR #72, opened 2020-07-15
Send credentials on requests to api

JackWilb:
Also fixed the string interpolation in this file

### DIFF
--- a/src/lib/multinet.ts
+++ b/src/lib/multinet.ts
@@ -1,5 +1,6 @@
 /* Multinet data importer */
 import { multinetApi } from 'multinet';
+import { Node, Network, Link } from '@/types';
 
 async function _downloadAllRows(api: any, workspace: string, tableName: string, tableType: 'node' | 'link') {
   let table = await api.table(workspace, tableName);
@@ -24,6 +25,7 @@ async function _downloadAllRows(api: any, workspace: string, tableName: string, 
   return output;
 }
 
+function _renameLinkVars(links: any[]): Link[] {
   for (const row of links) {
     row.id = row._id;
     row.source = row._from;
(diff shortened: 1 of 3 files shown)


src/lib/multinet.ts
```diff
@@ -1,24 +1,31 @@
...
-async function _loadLinks(workspace: string, edgeTable: string, apiRoot: string) {
-  const linksCall = apiRoot + '/workspaces/' + workspace + '/tables/' + edgeTable + '?limit=1000';
-  const linksRaw = await json(linksCall);
-  return linksRaw.rows;
+  // Else if the table is small enough, loop through the rows and download
+  let output: any[] = [];
+  output = output.concat(table.rows);
+
+  while (output.length < table.count) {
+    table  = await api.table(workspace, tableName, { offset: output.length, limit: 30 });
```
waxlamp, line 21: You can probably up this to 100 rather than 30.

src/lib/multinet.ts
```diff
@@ -42,35 +49,37 @@ function _renameNodeVars(nodes: any[]) {
...
-  // Loop through each node tables and fetch the nodes to global variables
-  for (const nodeTable of multinet.tables.nodeTables) {
-    const ntable = await _loadNodes(workspace, nodeTable, apiRoot);
-    multinet.nodes = multinet.nodes.concat(ntable);
+  // Loop through each node tables and fetch the nodes
+  for (const tableName of multinet.tables.nodeTables) {
+    multinet.nodes = multinet.nodes.concat(await _downloadAllRows(api, workspace, tableName, 'node'));
   }
 
   // Load the edge table (ONLY ONE BECAUSE OF ARANGO API LIMITATIONS)
```
waxlamp, line 81: This comment is a tiny bit misleading--it does originate from Arango's setup, but a single edge table is now part of the Multinet API as well. It's not a huge deal but I'd remove the parenthesis from this comment.

src/lib/multinet.ts
JackWilb: Fixed in 69b106e

src/lib/multinet.ts
JackWilb: Fixed in 69b106e

PR #77, opened 2020-08-07
Revised grid layout

mozartfish:
Tasks Completed (currently works for les mis and flight datasets)
- resized the cells to 11 x 11 
- resized grid for the new cells
- reshaped cells to have round corners
- fixed the row and column highlights to work with the new cell and grid size

Things that might need fixing
- resizing the labels (don't know if they are too big with the new layout)

### DIFF
--- a/src/components/AdjMatrix/AdjMatrixMethods.ts
+++ b/src/components/AdjMatrix/AdjMatrixMethods.ts
@@ -247,6 +247,12 @@ export class View {
     // set the radius for cells
     const cellRadius = 2;
 
+    // set the size of the number of nodes
+    const matrixNodeLength= this.network.nodes.length;
+
+    // set the matrix highlight
+    const matrixHighlightLength = matrixNodeLength * cellSize;
+
     // Use the smallest side as the length of the matrix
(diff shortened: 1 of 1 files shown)


src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -276,14 +282,17 @@ export class View {
 
     this.drawGridLines();
 
-    // add the highlight rows
+    // set the size of the highlight
+    const matrixHighlight = this.network.nodes.length * cellSize;
```
JackWilb, line 286: If you define this earlier you can reduce redundancy on 261 by using this var there

JackWilb: I think you should use a different, more descriptive, name here too. E.g. `matrixHighlightLength`

mozartfish: updated the code to reflect the changes you requested

mozartfish:
fixed the items that needed to be changed and started a new branch implementing Roni's suggestion

PR #78, opened 2020-08-10
Reshaping cells and removing grid

mozartfish:
This branch implements Roni's ideas about the UI with removing the gridlines and reshaping the filled cells so they do not touch. This branch also relies on the changes made in the revised-grid-layout branch
Selection_003

### DIFF
--- a/src/components/AdjMatrix/AdjMatrixMethods.ts
+++ b/src/components/AdjMatrix/AdjMatrixMethods.ts
@@ -247,6 +247,12 @@ export class View {
     // set the radius for cells
     const cellRadius = 2;
 
+    // set the size of the number of nodes
+    const matrixNodeLength= this.network.nodes.length;
+
+    // set the matrix highlight
+    const matrixHighlightLength = matrixNodeLength * cellSize;
+
     // Use the smallest side as the length of the matrix
     this.edgeWidth = sideLength - (this.margins.left + this.margins.right);
     this.edgeHeight = sideLength - (this.margins.top + this.margins.bottom);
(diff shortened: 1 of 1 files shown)


JackWilb:
What's going on with the highlight column in the image there?

mozartfish:
The highlight seems to appear when I hover over some rows like this 
Screenshot from 2020-08-10 13-27-48

mozartfish:
it might be a bug

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -309,8 +322,9 @@ export class View {
       .attr('class', 'cell')
       .attr('id', (d: Cell) => d.cellName)
       .attr('x', (d: Cell) => this.orderingScale(d.x))
-      .attr('width', this.orderingScale.bandwidth())
-      .attr('height', this.orderingScale.bandwidth())
+      .attr('width', cellSize - 2)
+      .attr('height', cellSize - 2)
```
waxlamp, line 326: You may want to abstract this out into a `const` variable above this block.

Can you try using `cellSize - 1`? The current look is a bit too dot-like.

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -548,35 +562,35 @@ export class View {
...
+    //   .style('opacity', 0.3);
+
+    // gridLines
+    //   .append('line')
+    //   .attr('x1', 0)
+    //   .attr('x2', this.orderingScale.range()[1])
+    //   .attr('y1', this.orderingScale.range()[1])
+    //   .attr('y2', this.orderingScale.range()[1])
+    //   .style('stroke', '#aaa')
+    //   .style('opacity', 0.3);
```
waxlamp, line 593: Before you merge, you should either delete this code, or hide it behind an option of some sort.

I think having an option to show/hide the grid lines is actually preferable. But if that's too much work for this branch, I would say to probably uncomment this code and leave it in for the time being.

mozartfish: like a button event?

waxlamp: Probably a checkbox.

PR #83, opened 2020-08-13
Fixing labels

mozartfish:
Tasks Done
-Fixed the overlapping text by readjusting the font size
-Clipped the text so users can see the node name until the first grid line

Things that might need adjusting
- size of the text
- adding indicator for telling users that text got clipped
- resizing the grid to fit text

Selection_016
Selection_017


src/components/AdjMatrix/AdjMatrix.vue
```diff
@@ -158,6 +158,7 @@ svg >>> .colLabel,
 svg >>> .rowLabel {
   cursor: pointer;
   fill: black !important;
+
```
JackWilb, line 161:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -286,6 +286,7 @@ export class View {
       });
 
 
+
```
JackWilb, line 289:
```suggestion
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -472,14 +473,28 @@ export class View {
...
+    let clipShape = select('#text-clip').append('rect')
+    .attr('x', '-76')
+    .attr('y', '-5')
+    .attr('width', '60')
+    .attr('height', '100')
+
+
+
     this.edgeRows.append('text')
```
JackWilb, line 489:
```suggestion
    .attr('height', '100')

    this.edgeRows.append('text')
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -472,14 +473,28 @@ export class View {
    * @return none
    */
   private appendEdgeLabels(): void {
+    // add a clip path for the row labels
+    let textClip = select('#matrix');
+    textClip.append('clipPath')
+    .attr('id', 'text-clip');
+
+    let clipShape = select('#text-clip').append('rect')
+    .attr('x', '-76')
```
JackWilb, line 482: Why is this 76?

src/components/AdjMatrix/AdjMatrix.vue
mozartfish: So for adding the padding I added this as an experiment
Selection_018

but it seems to do nothing based on the output 
Selection_019

am I missing something?

JackWilb: Yes, the text is placed manually and adding padding has no effect on an svg

src/components/AdjMatrix/AdjMatrixMethods.ts
mozartfish: This is 76 because it was the x coordinate that allowed for the clip window to display all the row label text without the row labels being cut off on the left and overlapping with the grid. If you have data with really long labels, then I can try to see if this value works for that.

mozartfish:
this is the new spacing for the labels where the labels do not bump into one another. There's an issue though with the amount of space needed to display the labels with this spacing. Any advice about what might be a potential fix (cut off some of the labels, smaller gap between nodes) would be greatly appreciated. 
Selection_022

mozartfish:
there were issues with saving the changes. I'm creating a new pull request

PR #84, opened 2020-08-14
Fixing labels

mozartfish:
Tasks Done
- resized the text with padding so there is no overlap with labels
- adjusted the svg so that all labels are displayed
- adjusted the highlights and the grid to accommodate the text changes
- rounded the corners of filled cells to a size that works for the grid
Selection_029
Selection_030


src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -472,14 +473,30 @@ export class View {
...
+    // add a clip path for the row labels
+    let textClip = select('#matrix');
+    textClip.append('clipPath')
+    .attr('id', 'text-clip');
+
+    let clipShape = select('#text-clip').append('rect')
+    .attr('x', '-80')
+    .attr('y', '-10')
+    .attr('width', '71')
+    .attr('height', '600')
```
JackWilb, line 485: This seem kind of arbitrary, are these based on any other values? e.g. the 71 = 74 - 3. If so, we could get that 74 into a variable and use that here too

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -294,9 +295,9 @@ export class View {
       .append('rect')
       .classed('topoCol', true)
       .attr('id', (d: Node) => `topoCol${d.id}`)
-      .attr('x', -matrixHighlightLength-this.margins.bottom)
+      .attr('x', -matrixHighlightLength)
       .attr('y', 0)
-      .attr('width', matrixHighlightLength + this.margins.top + this.margins.bottom)
+      .attr('width', matrixHighlightLength + this.margins.top + this.margins.bottom + 100)
```
JackWilb, line 300: Why did you have to increase the width? The row should already have been long enough for the label

JackWilb:
There are a couple other issues that I've spotted. The cells render in the top right of the gridlines (bump x by 1 and y by 1) and the full matrix doesn't render on first load (stop the auto sizing based on screen size and just use a fixed value for width for now)

src/components/AdjMatrix/AdjMatrixMethods.ts
mozartfish: That was a bug I introduced and removed it with a change

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -472,14 +473,30 @@ export class View {
...
     this.edgeRows.append('text')
+    .attr('clip-path', 'url(#text-clip)')
       .attr('class', 'rowLabel')
       .attr('id', (d: Node) => `rowLabel${d.id}`)
       .attr('z-index', 30)
-      .attr('x', -3)
-      .attr('y', this.orderingScale.bandwidth() / 2)
-      .attr('dy', '.32em')
-      .attr('text-anchor', 'end')
+      .attr('x', -74)
```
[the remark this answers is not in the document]
mozartfish: I saw that for some of the labels for the Les Mis Dataset that letters such as C, G, and others were being cut off because of the clipping. By changing the x coordinate to 74, I found a way to make sure that none of the labels got cut off when they were were drawn. I do agree that this value could be put in a const so that in further iterations if we need to fix the text we can update the const.

src/components/AdjMatrix/AdjMatrix.vue
```diff
@@ -86,13 +86,9 @@ export default Vue.extend({
...
   async mounted(this: any) {
-    this.browser.width = window.innerWidth
-      || document.documentElement.clientWidth
-      || document.body.clientWidth;
+    this.browser.width = 100000
 
-    this.browser.height = window.innerHeight
-      || document.documentElement.clientHeight
-      || document.body.clientHeight;
+    this.browser.height = 10000
```
JackWilb, line 91: We know how big the matrix needs to be based on the size of the network * cell size. Let's use that instead of making a massive vertical scroll

PR #86, opened 2020-08-19
use .env/env-vars to set multinet api host

### LINKED ISSUE #76: Use an environment variable to set the Multinet host at build time

JackWilb:
Depends on #85

Closes #76 

You can test each file with `yarn serve` and `yarn serve --mode=production`

`yarn serve` defaults to development mode
`yarn build` defaults to production mode

[Reference](https://cli.vuejs.org/guide/mode-and-env.html#modes)

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+VUE_APP_MULTINET_API_URL=https://api.multinet.app/api
--- a/.env
+++ b/.env
@@ -1,1 +1,0 @@
-VUE_APP_MULTINET_API_URL=https://api.multinet.app/api
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ To get the application up and running in dev mode:
 
 1. Ensure that you have Node and yarn installed.
 2. Check out this repository and move into it (e.g., `cd view-adjmatrix`).
-3. Copy the correct .env.development to .env with `cp .env.development .env`
+3. Copy .env.development to .env with `cp .env.development .env`
 4. Install the Yarn dependencies: `yarn install`.
 5. Run the application: `yarn serve`.
 
--- a/README.md
+++ b/README.md
@@ -10,8 +10,9 @@ To get the application up and running in dev mode:
 
 1. Ensure that you have Node and yarn installed.
 2. Check out this repository and move into it (e.g., `cd view-adjmatrix`).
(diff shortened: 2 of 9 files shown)


.env
```diff
@@ -0,0 +1 @@
+VUE_APP_MULTINET_API_URL=https://api.multinet.app/api
```
waxlamp, line 1: For consistency with the client, this should be `VUE_APP_MULTINET_HOST`.

JackWilb:
I think this would work without the netlify.toml, because the .env.production gets read automatically into the environment when the mode is production, which it is when you run `yarn build`. Not sure if you'd like to delete it, for now. We could also remove the .env.production in favor of the netlify.toml

waxlamp:
> I think this would work without the netlify.toml, because the .env.production gets read automatically into the environment when the mode is production, which it is when you run `yarn build`. Not sure if you'd like to delete it, for now. We could also remove the .env.production in favor of the netlify.toml

Is this something `yarn` does automatically? The [Netlify docs on environment variables](https://docs.netlify.com/configure-builds/environment-variables/#declare-variables) don't mention `.env.*` files as a way to set these up. I think it would be better to use `netlify.toml` to control this, since it also allows you to tune variables and other settings based not just on build-vs-dev but also on deploy context.

So I would say let's get rid of `.env.production` and let `netlify.toml` be the mechanism that controls how Netlify builds things.

JackWilb:
> Is this something `yarn` does automatically? T

Yes it is.

> So I would say let's get rid of `.env.production` and let `netlify.toml` be the mechanism that controls how Netlify builds things.

Agreed, see 886a351

README.md
```diff
@@ -10,8 +10,9 @@ To get the application up and running in dev mode:
 
 1. Ensure that you have Node and yarn installed.
 2. Check out this repository and move into it (e.g., `cd view-adjmatrix`).
-3. Install the Yarn dependencies: `yarn install`.
-4. Run the application: `yarn serve`.
+3. Copy the correct .env.development to .env with `cp .env.development .env`
```
waxlamp, line 13:
```suggestion
3. Copy .env.development to .env with `cp .env.development .env`
```

waxlamp (dismissed):
One suggested change; otherwise looks good.

PR #88, opened 2020-08-24
Fixing labels revision1

mozartfish:
Fixed everything in the fix labels branch before the highlight broke and the rendering problems did not work. 
The changes made were all of the changes requested in the fixed-labels PR request which I have reopened. The code in that request is broken so this is the code that needs to be reviewed. The issue that hasn't been fixed (which I am still having problems with) is resizing the svgs without the highlights crashing and having the svg work for any size network data.

### DIFF
--- a/src/components/AdjMatrix/AdjMatrixMethods.ts
+++ b/src/components/AdjMatrix/AdjMatrixMethods.ts
@@ -586,14 +586,14 @@ export class View {
       .data(this.matrix)
       .enter();
 
-      //vertical grid lines
+    //vertical grid lines
     lines.append('line')
       .attr('transform', (d: any, i: number) => {
         return `translate(${this.orderingScale(i)},0)rotate(-90)`;
       })
       .attr('x1', -this.orderingScale.range()[1]);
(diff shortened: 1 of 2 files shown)


src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -531,7 +539,7 @@ export class View {
       .attr('x', verticalOffset)
       .attr('dy', '.32em')
       .attr('text-anchor', 'start')
-      .style('font-size', this.nodeFontSize)
+      .style('font-size', this.nodeFontSize.toString() + 'pt')
```
JackWilb, line 542: This is already a string so we can just do (there are 2 places to change this):
```suggestion
      .style('font-size', `${this.nodeFontSize}pt`)
```

JackWilb (changes requested):
One super small suggestion

JackWilb (approved):
Looks great. Feel free to take this out of draft mode, merge this, and close your other open PRs related to this change

PR #93, opened 2020-09-01
Implement ellipsis for long label cutoff for the adjacency matrix

mozartfish:
This prototype is a solution for how to handle the label cutoff problem with SVG text using SVG Foreign Objects.

This is a picture with the foreign objects for the rows 
image

currently working on foreign objects for columns

### DIFF
--- a/src/components/AdjMatrix/AdjMatrixMethods.ts
+++ b/src/components/AdjMatrix/AdjMatrixMethods.ts
@@ -561,17 +561,20 @@ export class View {
 
     verticalOffset = verticalOffset * 0.075 + 5;
 
-    this.edgeColumns
-      .append('text')
-      .attr('id', (d: Node) => `colLabel${d.id}`)
-      .attr('class', 'colLabel')
-      .attr('z-index', 30)
-      .attr('y', this.orderingScale.bandwidth() / 2)
-      .attr('x', verticalOffset)
-      .attr('dy', '.32em')
-      .attr('text-anchor', 'start')
-      .style('font-size', `${this.nodeFontSize}pt`)
+    // constant for starting the column label container
+    const columnLabelContainerStart = 20;
+
+    const edgeColumnForeignObject = this.edgeColumns
+      .append('foreignObject')
+      .attr('y', -5)
+      .attr('x', columnLabelContainerStart)
+      .attr('width', labelContainerWidth)
+      .attr('height', labelContainerHeight);
+
+    const colLabel = edgeColumnForeignObject
+      .append('xhtml:p')
       .text((d: Node) => d._key)
+      .classed('colLabels', true)
       .on('click', (d: Node) => {
         this.selectElement(d);
         this.selectNeighborNodes(d.id, d.neighbors);
--- a/src/components/AdjMatrix/AdjMatrixMethods.ts
+++ b/src/components/AdjMatrix/AdjMatrixMethods.ts
@@ -505,39 +505,22 @@ export class View {
    * @return none
    */
   private appendEdgeLabels(): void {
-    // constants for manipulating the x and y values of the different clip parts
-    const clipXValue = 74;
-    const clipYValue = 10;
+    const labelContainerHeight = 25;
+    const rowLabelContainerStart = 75;
+    const labelContainerWidth = rowLabelContainerStart;
 
-    // clip path for row labels
-    select('#matrix').append('clipPath').attr('id', 'text-clip');
+    // add foreign objects for label
+    const edgeRowForeignObject = this.edgeRows
+      .append('foreignObject')
+      .attr('x', -rowLabelContainerStart)
+      .attr('y', -5)
+      .attr('width', labelContainerWidth)
+      .attr('height', labelContainerHeight);
 
-    // the shape of the clip path
-    select('#text-clip')
-      .append('rect')
-      .attr('x', -(clipXValue + 6))
-      .attr('y', -clipYValue)
-      .attr('width', clipXValue)
-      .attr('height', 600);
-
-    this.edgeRows
-      .append('text')
-      .attr('clip-path', 'url(#text-clip)')
-      .attr('class', 'rowLabel')
-      .attr('id', (d: Node) => `rowLabel${d.id}`)
-      .attr('z-index', 30)
-      .attr('x', -clipXValue)
-      .attr('y', clipYValue / 2)
-      .attr('dy', '.75em')
-      .attr('text-anchor', 'start')
-      .style('font-size', `${this.nodeFontSize}pt`)
+    const rowLabel = edgeRowForeignObject
+      .append('xhtml:p')
       .text((d: Node) => d._key)
-      .on('mouseover', (d: Node) => this.hoverNode(d.id))
-      .on('mouseout', (d: Node) => this.unHoverNode(d.id))
-      .on('mouseover', (d: Node, i: number, nodes: any) => {
-        this.showToolTip(d, i, nodes);
-        this.hoverNode(d.id);
-      })
+      .classed('rowLabels', true)
       .on('mouseout', (d: Node) => {
         this.hideToolTip();
         this.unHoverNode(d.id);
--- a/src/components/AdjMatrix/AdjMatrix.vue
+++ b/src/components/AdjMatrix/AdjMatrix.vue
@@ -144,6 +144,22 @@ svg >>> .baseCell {
   fill-opacity: 0;
 }
 
+svg >>> .rowLabels {
+  max-width: 75px;
+  text-overflow: ellipsis;
+  overflow: hidden;
+  font-size: 12pt;
+  z-index: 100;
+}
+
+svg >>> .colLabels {
+  max-width: 55px;
+  text-overflow: ellipsis;
+  overflow: hidden;
+  font-size: 12pt;
+  z-index: 100;
+}
+
 svg >>> .hoveredCell {
   stroke-width: 1px;
   stroke: darkgray;


mozartfish:
Finished the row and column labels for the matrix.

Les Mis
image

Flight data 
image

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -505,39 +505,28 @@ export class View {
...
-        this.showToolTip(d, i, nodes);
-        this.hoverNode(d.id);
-      })
+    // add foreign objects for label
+    const edgeRowForeignObject = this.edgeRows
+      .append('foreignObject')
+      .attr('x', -foreignObjectXConst)
+      .attr('y', 0)
+      .attr('width', foreignObjectDimensionsConstant + 60)
+      .attr('height', foreignObjectDimensionsConstant);
```
waxlamp, line 517: Since the width and height for these are set, and also used in the column labels below, I think you'd be better off making both of these `const` values that are more explicitly set. It also becomes easier to name these variables (`labelWidth` and `labelHeight`, or something like that).

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -578,17 +567,20 @@ export class View {
...
-      .attr('y', this.orderingScale.bandwidth() / 2)
-      .attr('x', verticalOffset)
-      .attr('dy', '.32em')
-      .attr('text-anchor', 'start')
-      .style('font-size', `${this.nodeFontSize}pt`)
-      .text((d: Node) => d._key)
+    const edgeColumnForeignObject = this.edgeColumns
+      .append('foreignObject')
+      .attr('y', 0)
+      .attr('x', -foreignObjectXConst + 95)
```
waxlamp, line 573: This `95` is a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) so maybe we can define the full expression as a constant, or at least give the `95` a name via a `const` variable.

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -578,17 +567,20 @@ export class View {
...
+      .attr('width', foreignObjectDimensionsConstant + 60)
+      .attr('height', foreignObjectDimensionsConstant);
```
waxlamp, line 575: See above comment; whatever solution is used there can be deployed in this instance as well.

waxlamp (changes requested):
Looking good. I have several suggestions about how to improve the code, but the visuals are very nice.

Before merging:
- [ ] file an issue to hoist some of the DOM creation to the Vue layer

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -505,39 +505,28 @@ export class View {
...
+    const rowLabelBody = edgeRowForeignObject.append('xhtml:body');
+
+    const rowLabel = rowLabelBody
+      .append('p')
+      .text((d: Node, i: number) => d._key)
+      // .text((d:Node, i:number) => {
+      //   console.log(d._key)
+      //   return d._key;
+      // })
+      .attr('z-index', 100)
```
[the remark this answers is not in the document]
mozartfish: The index is needed to show the importance of the labels when the highlight happens because if there is no z index, the highlight (which is an svg element) covers the label (a paragraph element within an svg element). To make sure the label has priority has over the highlight when a user hovers over a row/column, I used a index so the label can be seen while a user is highlighting a row/column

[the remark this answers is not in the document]
mozartfish: Just realized that I could put that in the css

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -578,17 +567,20 @@ export class View {
...
+      .attr('x', -foreignObjectXConst + 95)
+      .attr('width', foreignObjectDimensionsConstant + 60)
+      .attr('height', foreignObjectDimensionsConstant);
+
+    const colLabelBody = edgeColumnForeignObject.append('xhtml:body');
+
+    const colLabel = colLabelBody
+      .append('p')
+      .text((d: Node, i: number) => d._key)
+      .attr('z-index', 100)
```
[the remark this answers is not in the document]
mozartfish: put the z index in the vue file

src/components/AdjMatrix/AdjMatrixMethods.ts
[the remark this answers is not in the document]
waxlamp: Whether this stays here or moves to CSS, it will be important to leave a comment explaining what this is being used for. In general, hard-coding values like this (even when it's necessary) can lead to confusion later if there's no attending explanation.

src/components/AdjMatrix/AdjMatrixMethods.ts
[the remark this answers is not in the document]
waxlamp: > put the z index in the vue file

Do you mean you'll move this to the `<style>` section in the .vue file?

src/components/AdjMatrix/AdjMatrixMethods.ts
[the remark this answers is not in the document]
mozartfish: I moved this code to the css in the vue file

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -505,39 +505,24 @@ export class View {
...
-      .text((d: Node) => d._key)
-      .on('mouseover', (d: Node) => this.hoverNode(d.id))
-      .on('mouseout', (d: Node) => this.unHoverNode(d.id))
-      .on('mouseover', (d: Node, i: number, nodes: any) => {
-        this.showToolTip(d, i, nodes);
-        this.hoverNode(d.id);
-      })
+    const labelContainerHeight = 25;
+    const rowLabelContainerStart = 75;
+    const labelContainerWidth = labelContainerHeight + 60;
```
JackWilb, line 510: Why is this 85 and not 75? You can remove this and just use `rowLabelContainerStart`. Since you're only moving the label 75 pixels left from the edge of the matrix, you shouldn't have the width of the element be 85.

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -505,39 +505,24 @@ export class View {
...
+      .append('foreignObject')
+      .attr('x', -rowLabelContainerStart)
+      .attr('y', -5)
+      .attr('width', labelContainerWidth)
+      .attr('height', labelContainerHeight);
+
+    const rowLabelBody = edgeRowForeignObject.append('xhtml:body');
+
+    const rowLabel = rowLabelBody
+      .append('p')
```
JackWilb, line 523: You don't need to add a body each time, you can just add the p
```suggestion
      .attr('height', labelContainerHeight);

    const rowLabel = edgeRowForeignObject
      .append('xhtml:p')
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -505,39 +505,24 @@ export class View {
...
+      .text((d: Node, i: number) => d._key)
```
JackWilb, line 524:
```suggestion
      .text((d: Node) => d._key)
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -578,17 +563,22 @@ export class View {
...
+      .attr('y', -5)
+      .attr('x', columnLabelContainerStart)
+      .attr('width', labelContainerWidth)
+      .attr('height', labelContainerHeight);
+
+    const colLabelBody = edgeColumnForeignObject.append('xhtml:body');
+
+    const colLabel = colLabelBody
+      .append('p')
+      .text((d: Node, i: number) => d._key)
```
JackWilb, line 580:
```suggestion
      .text((d: Node) => d._key)
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -578,17 +563,22 @@ export class View {
...
+      .append('foreignObject')
+      .attr('y', -5)
+      .attr('x', columnLabelContainerStart)
+      .attr('width', labelContainerWidth)
+      .attr('height', labelContainerHeight);
+
+    const colLabelBody = edgeColumnForeignObject.append('xhtml:body');
+
+    const colLabel = colLabelBody
+      .append('p')
```
JackWilb, line 579:
```suggestion
      .attr('height', labelContainerHeight);

    const colLabel = edgeColumnForeignObject
      .append('xhtml:p')
```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -578,17 +563,22 @@ export class View {
...
-      .attr('class', 'colLabel')
-      .attr('z-index', 30)
-      .attr('y', this.orderingScale.bandwidth() / 2)
-      .attr('x', verticalOffset)
-      .attr('dy', '.32em')
-      .attr('text-anchor', 'start')
-      .style('font-size', `${this.nodeFontSize}pt`)
-      .text((d: Node) => d._key)
+    // constant for starting the column label container
+    const columnLabelContainerStart = -rowLabelContainerStart + 95;
```
JackWilb, line 567: I don't think this columnLabel variable should be tied to the rowLabels at all. This is just `-75 + 95`, you could just make it 20 since it doesn't rely on the rowLabels

JackWilb (changes requested):
Here are a couple issues I saw. We should clean these up and then rebase the commits at our meeting tomorrow and then it should be good to go

JackWilb (approved):
Looks good to me. Thanks for putting up with the changes to the code format, general fixes, and rebasing!

waxlamp:
> Looking good. I have several suggestions about how to improve the code, but the visuals are very nice.
> 
> Before merging:
> 
> * [ ]  file an issue to hoist some of the DOM creation to the Vue layer
> * [x]  let's consider rebasing the commit stream to improve the commit messages and/or restructure some of the changes

I see the rebasing was done (nice work); @mozartfish, could you please file the issue I requested and link it here?

Nice job on this PR!

PR #98, opened 2020-09-03
Add eslint and remove tslint

### LINKED ISSUE #97: Migrate to es-lint from ts-lint
We currently use ts-lint for linting the codebase. You can verify this by running `yarn serve` and noticing it says `tslint 5.20.1`. ts-lint is deprecated and will stop receiving security updates in 2021 so we should move to es-lint, which should maintain the same functionality.

JackWilb:
Closes #97 

Add eslint configs and remove tslint configs. There may still be some remnants to clean out, but that will require some more investigation/help (@waxlamp)

From the issue: 
> ts-lint is deprecated and will stop receiving security updates in 2021 so we should move to es-lint, which should maintain the same functionality.


.eslintrc.js
```diff
@@ -0,0 +1,31 @@
...
+      files: [
+        '**/__tests__/*.{j,t}s?(x)',
+        '**/tests/unit/**/*.spec.{j,t}s?(x)',
+      ],
+      env: {
+        jest: true,
+      },
+    },
+  ],
+};
```
waxlamp: Why move this to its own file, rather than leave it in package.json?

package.json
```diff
@@ -51,33 +51,5 @@
...
-        "env": {
-          "jest": true
-        }
-      }
-    ]
-  },
-  "browserslist": [
-    "> 1%",
-    "last 2 versions"
-  ],
```
waxlamp, line 75: Why remove this part?

.eslintrc.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   rules: {
     'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
     'no-debugger': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
+    '@typescript-eslint/no-explicit-any': 'off',
```
waxlamp, line 19: Why? This is an excellent rule to have.

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -393,9 +393,7 @@ export class View {
         .append('rect')
         .classed(`nestedEdges nestedEdges${type}`, true)
         .attr('x', offset)
-        .attr('y', (d: any) => {
-          return offset;
-        })
+        .attr('y', offset)
```
waxlamp, line 396: Interesting; how did ESLint know that this has the same semantics in D3?

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -937,7 +935,7 @@ export class View {
       path
         .append('path')
         .attr('class', 'sortIcon')
-        .attr('d', (d: any) => {
+        .attr('d', (_d: any, i: number) => {
```
waxlamp, line 938: Do you know why the linter replaced `d` with `_d`? Is it because this argument never appears in the function body?

waxlamp (commented):
Could you edit the description to explain what's different and/or better about using ESLint?

Also, you mentioned there are some remnants to clean up--what did you mean?

package.json
JackWilb: Mistake on my part. I though it was in the .eslintrc.js file. I'll add it back

.eslintrc.js
JackWilb: I meant to open up an issue here. I was planning to clean this up in a follow up

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: I think it complained about the any (before I removed that rule), so I just cleaned it up to be more concise.

.eslintrc.js
JackWilb: I think it just helps us be a bit more explicit. I like that the package.json is more for versioning dependencies and less for project config. Let me know if you'd prefer that I move it back

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: Yeah, eslint didn't like the variable being declare and not used. The quickest work around I could think of was to throw the _ on there; I think that's what some other linters I've used would suggest.

JackWilb:
> Could you edit the description to explain what's different and/or better about using ESLint?

It's in the issue, but I'll quote it up top for ease of reviewing.

> Also, you mentioned there are some remnants to clean up--what did you mean?

I didn't remove any of the plugins, and I was concerned that means that tslint is still in there as a dependency. I feel like that could be cleaned up a bit, but I'd need to explore what can be removed.

src/components/AdjMatrix/AdjMatrixMethods.ts
waxlamp: > The quickest work around I could think of was to throw the _ on there; I think that's what some other linters I've used would suggest.

If you can confirm this as a proper solution, that would be great. We should be sharpening up our best practices by using linters, and so "quick workarounds" are things to be verified if we use them.

.eslintrc.js
waxlamp: No, that's fine, I just wanted to know what the rationale was.

.eslintrc.js
waxlamp: Sounds good. Please link the issue here when it's opened.

JackWilb: Opened #100

package.json
JackWilb: See c1399d0

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: Looks like this is the current guidance [link](https://github.com/eslint/eslint/issues/1494), but there aren't any solid stances on what this should look like. It's a bit of a grey area so the guidance is either "explicitly disable this rule for that line", or "use a workaround" such as what I've done here

.eslintrc.js
```diff
@@ -0,0 +1,32 @@
...
+  rules: {
+    'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
+    'no-debugger': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
+    '@typescript-eslint/no-explicit-any': 'off',
+  },
+  overrides: [
+    {
+      files: [
+        '**/__tests__/*.{j,t}s?(x)',
+        '**/tests/unit/**/*.spec.{j,t}s?(x)',
```
waxlamp, line 25: These will be far more readable in the long term if you just make them say `{js,ts,jsx}`.

src/components/AdjMatrix/AdjMatrixMethods.ts
waxlamp: Did you try using the `after-used` option? That seems like what we actually want here. The issue you linked has a confusing comment the option `after-used` (but I disagree that the documentation is wrong; or rather, the docs *now* seem to reflect how this option ought to work).

waxlamp (dismissed):

.eslintrc.js
JackWilb: See efb9212

src/components/AdjMatrix/AdjMatrixMethods.ts
JackWilb: Yes and it worked! See db1d4d6

waxlamp (approved):

PR #102, opened 2020-09-12
Implement Grid Toggle for Showing/Hiding Matrix Gridlines

mozartfish:
This branch creates the grid toggle for showing/hiding the grid lines for the matrix.

The following images displays different kinds of toggles for showing/hiding the gridlines. The toggle selected was the checkbox because it easy to understand how to use and connects well with the adjacency matrix.
image

PR #103, opened 2020-09-14
Change the lone v-switch to a v-checkbox

waxlamp:
After #102, which uses a checkbox instead of a switch, the remaining switch in the UI feels out of place, and a checkbox seems like the better choice.

This tiny PR changes that switch to a checkbox.

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -108,7 +108,7 @@ export default Vue.extend({
               "
             >
               Autoselect neighbors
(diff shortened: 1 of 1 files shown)


JackWilb (approved):

mozartfish:
Looks good. The checkbox is more intuitive for users especially for things like a click or a selection and the checkbox style fits really well with the matrix UI look

mozartfish (approved):

PR #104, opened 2020-09-16
Add authentication token receiving from the querystring

JackWilb:
Adds a method to grab the login token from the query string so that a user can view data from private workspaces. This function is slightly different to that created by @AlmightyYakob, since we're not using vue router on this project, and the api is only handled in one spot on first load -- it's not continually accessed, unlike the main client.

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -74,7 +74,6 @@ export default Vue.extend({
 
       if (result !== null) {
         const { index, 1: token } = result;
-        console.log(index, token);
 
         const newPath = window.location.href.slice(0, index);
         window.history.replaceState({}, window.document.title, newPath);
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -6,7 +6,7 @@ import { getUrlVars } from '@/lib/utils';
 import { loadData } from '@/lib/multinet';
 import { Network } from '@/types';
 
-const loginTokenRegex = /&loginToken=(\S+)/;
+const loginTokenRegex = /#loginToken=(\S+)/;
 export default Vue.extend({
   components: {
     AdjMatrix,
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -6,7 +6,7 @@ import { getUrlVars } from '@/lib/utils';
 import { loadData } from '@/lib/multinet';
 import { Network } from '@/types';
 
-const loginTokenRegex = /&token=(\S+)/;
+const loginTokenRegex = /&loginToken=(\S+)/;
 export default Vue.extend({
   components: {
     AdjMatrix,
@@ -50,9 +50,9 @@ export default Vue.extend({
         `Workspace and network must be set! workspace=${workspace} network=${networkName}`,
       );
     }
-    const token = this.checkUrlForLogin();
+    const loginToken = this.checkUrlForLogin();
 
-    this.network = await loadData(workspace, networkName, host, token);
+    this.network = await loadData(workspace, networkName, host, loginToken);
     this.workspace = workspace;
     this.networkName = networkName;
   },
@@ -69,7 +69,7 @@ export default Vue.extend({
       a.click();
     },
 
-    checkUrlForLogin(this: any): string {
+    checkUrlForLogin(this: any): string | null {
       const result = loginTokenRegex.exec(window.location.href);
 
       if (result !== null) {
@@ -80,7 +80,7 @@ export default Vue.extend({
         return token;
       }
 
-      return '';
+      return null;
     },
   },
   watch: {
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -6,6 +6,7 @@ import { getUrlVars } from '@/lib/utils';
 import { loadData } from '@/lib/multinet';
 import { Network } from '@/types';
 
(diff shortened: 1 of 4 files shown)


src/components/Controls.vue
```diff
@@ -6,6 +6,7 @@ import { getUrlVars } from '@/lib/utils';
 import { loadData } from '@/lib/multinet';
 import { Network } from '@/types';
 
+const loginTokenRegex = /&token=(\S+)/;
```
AlmightyYakob, line 9: See https://github.com/multinet-app/multinet-client/pull/124#discussion_r490283373

src/components/Controls.vue
```diff
@@ -65,6 +68,20 @@ export default Vue.extend({
...
+
+      if (result !== null) {
+        const { index, 1: token } = result;
+
+        const newPath = window.location.href.slice(0, index);
+        window.history.replaceState({}, window.document.title, newPath);
+        return token;
+      }
+
+      return '';
```
AlmightyYakob, line 83: I think it'd be better if the token was explicitly null when missing, only calling `setAuthToken` when it's not null, as there's really no point in setting the token to `''`.

src/lib/multinet.ts
```diff
@@ -85,6 +86,7 @@ export async function loadData(
   };
 
   const api = multinetApi(apiRoot);
+  api.setAuthToken(token);
```
AlmightyYakob, line 89: Related to my comment above regarding null token value, there would be a check here against that.

AlmightyYakob (changes requested):
Small things, otherwise good

src/components/Controls.vue
```diff
@@ -65,6 +68,20 @@ export default Vue.extend({
       a.download = `${this.networkName}.json`;
       a.click();
     },
+
+    checkUrlForLogin(this: any): string {
+      const result = loginTokenRegex.exec(window.location.href);
```
waxlamp, line 73: Why is the regex declared globally? The code would be more readable if it was just used here directly.

src/components/Controls.vue
```diff
@@ -65,6 +68,20 @@ export default Vue.extend({
...
+
+      if (result !== null) {
+        const { index, 1: token } = result;
+
+        const newPath = window.location.href.slice(0, index);
```
waxlamp, line 78: This is pretty brittle. If the login token argument isn't the last thing in the URL, you're going to throw away some parts of it by accident.

Maybe a better approach is the fragment-based one @AlmightyYakob used for the client. @AlmightyYakob, thoughts?

waxlamp (changes requested):

src/components/Controls.vue
AlmightyYakob: I don't see a reason not to declare it globally, otherwise the variable would need to be recreated here unnecessarily every time it's called.

AlmightyYakob: I guess moreover, that regex isn't dependent on anything in this function, and is completely constant.

src/components/Controls.vue
AlmightyYakob: This is the same implementation as what I've done in the client. I think it's reasonable to assume that the token will be the last fragment in the URL. However it wouldn't be too hard to remove just the matching section of the URL, leaving what's before and after.

AlmightyYakob: No you're correct, I didn't notice that. @JackWilb I would also say it should be in the fragment

src/components/Controls.vue
JackWilb: I was following the style from the client here. I can update if necessary

src/components/Controls.vue
JackWilb: See 2291479

src/components/Controls.vue
JackWilb: See 2291479

src/lib/multinet.ts
JackWilb: See 2291479

src/components/Controls.vue
JackWilb: See e2c732a

src/components/Controls.vue
AlmightyYakob: @waxlamp okay with leaving the regex declared globally?

waxlamp: It's ok to leave it for now, but this is a classic case of "globals are bad". For example, if you move this function to somewhere else, but you forget the regex, it will break. Anyone reading this function for the first time will wonder where the variable came from, and have to find it at the top of the file, etc. And the argument about recreating it needlessly on every invocation is an example of premature optimization--we don't call this function very often, and it's really the compiler's job to recognize that this value never changes and do that sort of constant-factoring for us, etc.

I'm ok with leaving this as-is mainly because we're planning to refactor this out into multinetjs sometime soon. But in general, I don't like having a core part of this function defined as a global.

waxlamp (approved):

src/components/Controls.vue
AlmightyYakob: Good point about portability, I hadn't thought about that. I agree that I don't think it matters for now.

AlmightyYakob (approved):

PR #105, opened 2020-09-22
Implement the legend for the matrix in the matrix controls panel

mozartfish:


src/components/Controls.vue
```diff
@@ -52,6 +55,34 @@ export default Vue.extend({
...
+
+    // build legend for matrix
+    const neighborsList: number[] = [];
+    // get a list of all the neighbors for each node
+    this.network.nodes.forEach((element) => {
+      neighborsList.push(element.neighbors.length);
+    });
+    // find the max and min neighbors
+    const maxNumConnectionsLegend = max(neighborsList);
+    const minNumConnections = min(neighborsList);
```
JackWilb, line 67: We should use the computed max number of connections per cell here

yarn-error.log
```diff
@@ -0,0 +1,10928 @@
+Arguments: 
+  /usr/bin/node /usr/share/yarn/bin/yarn.js add @types/d3-legend
```
JackWilb, line 2: Remove this whole file

src/components/Controls.vue
```diff
@@ -121,6 +155,20 @@ export default Vue.extend({
...
+              class="pb-0 px-0"
+              style="
+                display: flex;
+                align-items: center;
+                justify-content: space-between;
+              "
+            >
+              Matrix Legend
+              <!-- <v-checkbox class="ma-0" v-model="showGridLines" hide-details /> -->
+              <svg id="matrix-legend"></svg>
```
JackWilb, line 170:
```suggestion
              Matrix Legend
              <svg id="matrix-legend"></svg>
```

src/components/Controls.vue
```diff
@@ -52,6 +55,34 @@ export default Vue.extend({
...
+
+    const legendLinear = legendColor()
+      .shapeWidth(40)
+      .orient('horizontal')
+      .scale(cellColorScaleLegend);
+
+    legendSVG.select('.legendLinear').call(legendLinear);
+
+    // console.log(this.network.nodes);
   },
```
JackWilb, line 86:
```suggestion
    legendSVG.select('.legendLinear').call(legendLinear);
  },
```

src/components/Controls.vue
JackWilb, line 67: You should consider spacing out your comments more

JackWilb (changes requested):
Please remove the whole yarn-error.log file, it's making it look like there are huge changes here.

Also, I think we should extract this logic into its own function in `methods:` and then call it from `mounted()` to keep it a bit more organized

src/components/Controls.vue
```diff
@@ -65,6 +68,28 @@ export default Vue.extend({
       a.download = `${this.networkName}.json`;
       a.click();
     },
+    createLegend(maxConnections: number) {
+
+      // set up color scale (currently only supports continuous data)
```
JackWilb, line 73:
```suggestion
    createLegend(maxConnections: number) {
      // set up color scale (currently only supports continuous data)
```

src/components/Controls.vue
```diff
@@ -65,6 +68,28 @@ export default Vue.extend({
...
+      const cellColorScaleLegend = scaleLinear<string>()
+        .domain([0, maxConnections])
+        .range(['#feebe2', '#690000']);
```
JackWilb, line 76: Just a thought. Instead of passing up the maxNumConnections like you have now. How about if we just pass the scale up. Then we wouldn't need to define a new scale here and the colors would update if we changed them in the methods.ts file

src/components/Controls.vue
```diff
@@ -140,8 +181,10 @@ export default Vue.extend({
               selectNeighbors,
               showGridLines,
               visualizedAttributes,
+              matrixView,
             }"
```
JackWilb, line 185:
```suggestion
              visualizedAttributes,
            }"
```

JackWilb (changes requested):
This looks great Pranav, and could honestly be merged in as is. However, I think you could simplify the logic a bit if instead of emitting the `maxNumConnections`, you just emit the scale and then use that for building the legend

src/components/AdjMatrix/AdjMatrix.vue
```diff
@@ -118,6 +118,7 @@ export default Vue.extend({
       this.cellSize,
       this.visMargins,
     );
+    this.$emit('updateMaxNumConnections', this.view.colorScale);
```
JackWilb, line 121: Change the event name, `updateMaxNumConnections` isn't what we're doing anymore. We're passing the whole scale

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -8,6 +8,7 @@ import * as ProvenanceLibrary from 'provenance-lib-core/lib/src/provenance-core/
 import 'science';
 import 'reorder.js';
 import { Link, Network, Node, Cell, State } from '@/types';
+import { ScaleLinear } from 'd3';
```
JackWilb, line 12: This is already imported on line 2
```suggestion
import { Link, Network, Node, Cell, State } from '@/types';

```

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -333,9 +335,11 @@ export class View {
       .attr('height', this.orderingScale.bandwidth())
       .attr('fill-opacity', 0);
 
-    const cellColorScale = scaleLinear<string>()
-      .domain([0, this.maxNumConnections])
-      .range(['#feebe2', '#690000']); // TODO: colors here are arbitrary, change later
+    // const cellColorScale = scaleLinear<string>()
+    //   .domain([0, this.maxNumConnections])
+    //   .range(['#feebe2', '#690000']); // TODO: colors here are arbitrary, change later
```
JackWilb, line 340:
```suggestion
```

src/components/Controls.vue
```diff
@@ -142,6 +178,7 @@ export default Vue.extend({
               visualizedAttributes,
             }"
             @restart-simulation="hello()"
+            @updateMaxNumConnections="createLegend"
```
JackWilb, line 181: When you update the name of the event in the emit, it will need to be changed here, too.

src/components/Controls.vue
```diff
@@ -1,7 +1,10 @@
 <script lang="ts">
 import Vue from 'vue';
 import AdjMatrix from '@/components/AdjMatrix/AdjMatrix.vue';
-import { selectAll } from 'd3-selection';
+import { select, selectAll } from 'd3-selection';
+import { format } from 'd3-format';
+import { legendColor } from 'd3-svg-legend';
+import { ScaleLinear,} from 'd3-scale';
```
JackWilb, line 7:
```suggestion
import { ScaleLinear } from 'd3-scale';
```

src/components/Controls.vue
```diff
@@ -65,6 +68,23 @@ export default Vue.extend({
       a.download = `${this.networkName}.json`;
       a.click();
     },
+    createLegend(colorScale: ScaleLinear<string, number>) {
+
+      const legendSVG = select('#matrix-legend');
```
JackWilb, line 73:
```suggestion
    createLegend(colorScale: ScaleLinear<string, number>) {
      const legendSVG = select('#matrix-legend');
```

JackWilb (changes requested):
Ping me when you're done with these changes and let's see if we can fix these typescript errors on the build

waxlamp (changes requested):
This looks good at a glance. One thing to fix is that you seem to have checked in a yarn error file; please make sure that's removed from source control.

The only other comment I have is that the color legend should *probably* be its own Vue component. But that's something we can address later.

JackWilb:
> This looks good at a glance. One thing to fix is that you seem to have checked in a yarn error file; please make sure that's removed from source control.

@mozartfish You should make a commit that removes the file and adds it to the .gitignore so it doesn't keep getting accidentally added.

waxlamp:
> > This looks good at a glance. One thing to fix is that you seem to have checked in a yarn error file; please make sure that's removed from source control.
> 
> @mozartfish You should make a commit that removes the file and adds it to the .gitignore so it doesn't keep getting accidentally added.

The other thing is: don't use commands like `git add *` (not sure if this is what happened); instead, be very explicit about adding files.

.gitignore
```diff
@@ -11,4 +11,5 @@ dist/
 .env
 
 # yarn error log file
-yarn-error.log
\ No newline at end of file
+yarn-error.log
+
```
waxlamp, line 15:
```suggestion
```

waxlamp (approved):
There is a blank line at the end of a file that you should strike, but otherwise this looks good. Thanks!

JackWilb (approved):

PR #109, opened 2020-10-14
Update @vue/cli-service to fix vulns

JackWilb:
node-forge had vulnerabilities in 0.9.0. The package was a dependency of selfsigned --> webpack-dev-server--> @vue/cli-service.

Thus, I ran yarn upgrade @vue/cli-service and this is the resulting lock file changes.

This also cleans up the issue with `serialize-javascript`, which was a deep dependency of @vue/cli-service

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,7 +2,14 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.8.3":
+"@babel/code-frame@^7.0.0":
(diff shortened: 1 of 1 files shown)


waxlamp (changes requested):
If you can update this to match the similar fix from multinet-client, I can approve.

You can use this as a chance to test out the auto-updating capability of Yarn on the package URLs (i.e., move your `node_modules` directory to a temp name and then re-run `yarn install` with an updated lock file).

PR #110, opened 2020-10-16
State Aggregation Function

mozartfish:
This PR attempts to create a supernode graph for a small adjacency matrix


src/components/Controls.vue
```diff
@@ -8,8 +8,72 @@ import { ScaleLinear } from 'd3-scale';
...
 import { loadData } from '@/lib/multinet';
 import { Network } from '@/types';
+// import { View } from './AdjMatrix/AdjMatrixMethods';
 
 const loginTokenRegex = /#loginToken=(\S+)/;
+
+/**
+ * This function takes a list of nodes and identifies a supernode
+ * and returns a new super node and a list of edges associated with the supernode
+ */
```
waxlamp, line 18: Just as a general matter of style, we use `// ` prefixed lines even for comment blocks like this.

src/components/Controls.vue
```diff
@@ -8,8 +8,72 @@ import { ScaleLinear } from 'd3-scale';
...
+  );
+
+  console.log('california nodes');
+  console.log(californiaNodes);
+
+  // modify all the california node id and keys
+  californiaNodes.forEach((node) => {
+    node['_key'] = 'CA';
+    node['id'] = 'nodes/CA';
+  });
```
waxlamp, line 44: I don't think this is right: this just converts all California nodes to have the same key and ID, rather than creating a supernode that represents all such nodes.

src/components/Controls.vue
```diff
@@ -8,8 +8,72 @@ import { ScaleLinear } from 'd3-scale';
...
+
+  // modify all the neighbors of the other nodes to reflect the california super node change
+  otherNodes.forEach((node) => {
+    const nodeNeighbors = node['neighbors'];
+    for (let i = 0; i < nodeNeighbors.length; i++) {
+      if (
+        nodeNeighbors[i] === 'nodes/SFO' ||
+        nodeNeighbors[i] === 'nodes/OAK'
```
waxlamp, line 52: Why not check to see if the state of the airport is California?

src/components/Controls.vue
```diff
@@ -8,8 +8,72 @@ import { ScaleLinear } from 'd3-scale';
...
+    if (edge['source'] === 'nodes/SFO' || edge['source'] === 'nodes/OAK') {
+      edge['source'] = 'nodes/CA';
+    }
+    if (edge['target'] === 'nodes/SFO' || edge['target'] === 'nodes/OAK') {
+      edge['target'] = 'nodes/CA';
+    }
+  });
+  console.log('new edges');
+
+  const newNodes = otherNodes.concat(californiaNodes);
```
waxlamp, line 71: This isn't quite the approach you should take. You have in-place modified the original data, and then you've made a copy. I believe a better approach would be to perform a `forEach()` over the original nodes, and generate supernodes as needed to refer to the original nodes. Then you'll have both the (unmodified) original nodes, and a new list of supernodes. You will want to use a name like "supernodes" for the nodes in this new list. Finally, you'll want to do an `edges.forEach()` (as you did in line 61ff. above), but generate new edges there as well. You should under no circumstances be modifying the original data.

Once you have a list of supernodes and a list of new edges, you will have what you need for a transformed graph.

This is fairly complex and I may not have explained it well so we should meet in person to discuss it. @gotdairyya should be there as well.

src/components/Controls.vue
```diff
@@ -1,10 +1,151 @@
...
+  const newNodes: any[] = [];
+  nodes.forEach((node: any) => {
+    const newNode: any = {};
+    newNode['ORIGIN'] = node['ORIGIN'];
+    newNode['ORIGIN_CITY'] = node['ORIGIN_CITY'];
+    newNode['ORIGIN_STATE'] = node['ORIGIN_STATE'];
+    newNode['_key'] = node['_key'];
+    newNode['_id'] = node['id'];
+    newNode['parent'] = undefined;
+    newNodes.push(newNode);
```
waxlamp, line 34: There's an easier way to express this in ES6. It will be something like:

```typescript
const newNode = Object.assign({}, node, {
  parent: undefined,
});
```

Even better is the spread operator:

```typescript
const newNode = {
  ...node,
  parent: undefined,
};
```

You may want to read some documentation and experiment with these forms to make sure they work the way you expect.

waxlamp: Some more comments about the code itself.

1. What is the purpose of the `parent` property?
2. You may not want to copy over the `_key` and `_id` fields exactly. If you wanted to push the new nodes list back to Multinet, you'd end up causing a problem, since nodes with those ID values already exist in the database. In general, I don't think it's necessary to copy every node like this; instead, you can simply reuse the original nodes in specifying the new network. It may require a meeting to discuss this, because it's sort of an area we have not explored too much yet.

src/components/Controls.vue
```diff
@@ -1,10 +1,151 @@
...
+
+  // find the index of the super node
+  const superIndex = newNodes.findIndex(
+    (node: any) => node['_id'] === 'supernodes/CA',
+  );
+
+  newNodes.forEach((node) => {
+    if (
+      node['_id'] !== 'supernodes/CA' &&
+      node['ORIGIN_STATE'] === 'California'
```
waxlamp, line 61: Any time you are referencing a JS object's properties statically like this, it's preferred to write it as `node._id` or `node.ORIGIN_STATE`, etc.

src/components/Controls.vue
```diff
@@ -1,10 +1,151 @@
...
+    ) {
+      node['parent'] = superIndex;
+      newNodes[superIndex]['children'].push(node['index']);
```
waxlamp, line 64: Instead of recording the index in the children list, you should record the child node's ID. This will be unique within a Multinet instance, and is the canonical way to reference other nodes/edges in the system.

src/components/Controls.vue
```diff
@@ -1,10 +1,151 @@
...
+    newLink['AIR_TIME_MEDIAN'] = link['AIR_TIME_MEDIAN'];
+    newLink['AIR_TIME_MEAN'] = link['AIR_TIME_MEAN'];
+    newLink['DEP_DELAY_MAX'] = link['DEP_DELAY_MAX'];
+    newLink['AIR_TIME_MAX'] = link['AIR_TIME_MAX'];
+    newLink['AIRLINE'] = link['AIRLINE'];
+    newLink['AIR_TIME_MIN'] = link['AIR_TIME_MIN'];
+    newLink['DEP_DELAY_MEDIAN'] = link['DEP_DELAY_MEDIAN'];
+    newLink['DEP_DELAY_MEAN'] = link['DEP_DELAY_MEAN'];
+    newLink['DEP_DELAY_MIN'] = link['DEP_DELAY_MIN'];
+    newLinks.push(newLink);
```
waxlamp, line 96: Use `Object.assign()` or (better) the spread operator here, as mentioned above.

src/components/Controls.vue
```diff
@@ -1,10 +1,151 @@
...
+  });
+
+  // update all the source and destinations
+  newLinks.forEach((link: any) => {
+    const linkFrom = link['_from'];
+    const linkTo = link['_to'];
+    if (newNodeDict[linkFrom] != undefined) {
+      const index = newNodeDict[linkFrom];
+      const newLinkFrom = newNodes[index]['_id'];
+      link['_from'] = newLinkFrom;
```
waxlamp, line 106: The `_from` and `_to` properties MUST be ArangoDB ID values, NOT index values. This is a similar comment as above, but in some sense more important here, since ArangoDB (and thus Multinet) uses these fields to perform graph-based searches, etc.

src/components/Controls.vue
```diff
@@ -1,10 +1,151 @@
...
+    }
+    if (newNodeDict[linkTo] != undefined) {
+      const index = newNodeDict[linkTo];
+      const newLinkTo = newNodes[index]['_id'];
+      link['_to'] = newLinkTo;
+    }
+  });
+
+  // update the edge id of the links
+  for (let index = 0; index < newLinks.length; index++) {
```
waxlamp, line 116: Rather than a for loop, you should use `Array.forEach()`. The second argument to `forEach()` is the index of each item. So basically, it's something like:

```typescript
newLinks.forEach((link, index) => {
  .
  .
  .
});
```

src/components/Controls.vue
```diff
@@ -1,10 +1,151 @@
...
+    const edgeName = 'edges/' + index;
```
waxlamp, line 117: Use a template literal here:

```suggestion
    const edgeName = `edges/${index}`;
```

waxlamp (commented):
This is great progress, @mozartfish! I think we may need to talk about how to improve this further. But it's a great milestone that this actually works end-to-end within the application.

waxlamp (commented):
I think this is looking pretty good, @pranav! This branch has become rather messy because of a few factors, so let's spend some effort in cleaning it up (and perhaps rebasing the commits) and then I believe it will be ready for a final review and merge.

As part of that review, I think we'll have uncovered some problems with the architecture of the main application. If we make a note of these, we can fix those problems in follow-up PRs. (The main problem is, I believe, that the application mutates the input data in place to prep it for D3, and never expects that input data to change. I'm picturing more of a pipeline built around computed props, etc., that would allow the input data to change with minimal disruption to the D3 portion of the app. We'll need to discuss this a bit before we know what to do. cc @JackWilb)

src/lib/multinet.ts
```diff
@@ -32,23 +32,25 @@ async function _downloadAllRows(
     });
     output = output.concat(table.rows);
   }
-
   return output;
 }
 
-function _renameLinkVars(links: any[]): Link[] {
+// Function to refactor link properties to visualize links with d3
+// keep the _from and _to fields to make it easier to folllow the multinet data format
```
JackWilb, line 39: Remove line 39

src/lib/multinet.ts
```diff
@@ -32,23 +32,25 @@ async function _downloadAllRows(
...
   for (const row of links) {
     row.id = row._id;
     row.source = row._from;
     row.target = row._to;
 
     delete row._id;
-    delete row._from;
-    delete row._to;
+    // delete row._from;
+    // delete row._to;
```
JackWilb, line 48: Delete these

src/lib/multinet.ts
```diff
@@ -66,6 +69,16 @@ function _defineNeighbors(nodes: any[], links: any[]) {
...
+// Function that constructs the neighbors for a node in a super network
+export function defineSuperNeighbors(nodes: any[], links: any[]) {
+  nodes.map((d: { neighbors: string[] }) => (d.neighbors = []));
+  for (const link of links) {
+    nodes.filter((d: Node) => d.id === link._from)[0].neighbors.push(link._to);
+    nodes.filter((d: Node) => d.id === link._to)[0].neighbors.push(link._from);
+  }
+  return nodes;
+}
+
```
JackWilb, line 81: Since this is the exact same logic as above, you can remove this and use the other function. You'll just want to export the function above and remove the `_` from the front

src/lib/multinet.ts
```diff
@@ -32,23 +32,25 @@ async function _downloadAllRows(
     });
     output = output.concat(table.rows);
   }
-
```
JackWilb, line 35: Was this a linter suggestion? Otherwise, we can leave the line there

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
 import { legendColor } from 'd3-svg-legend';
 import { ScaleLinear } from 'd3-scale';
 import { getUrlVars } from '@/lib/utils';
-import { loadData } from '@/lib/multinet';
+import { loadData, defineSuperNeighbors } from '@/lib/multinet';
```
JackWilb, line 9: Once you remove this function in favor of `defineNeighbors`, you'll want to update the import here

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
 import { Network } from '@/types';
+// import { View } from './AdjMatrix/AdjMatrixMethods';
```
JackWilb, line 11: Looks like this remnant can be removed

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+// and creates a new list of supernodes and a new list of edges
+// to reflect the connections between a supernode and the original nodes
+// in the network
+function superGraph(nodes: any[], edges: any[]) {
+  // print the original nodes and edges
+  // console.log('THE ORIGINAL NODES');
+  // console.log(nodes);
+  // console.log('THE ORIGINAL EDGES');
+  // console.log(edges);
+
```
JackWilb, line 23: Can be removed

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+    const newNode = {
+      ...node,
+      // add new attributes for the new nodes
+      // index - attribute for keeping track of the index for visualizing the network
+      index: index,
+    };
+
+    // remove the properties that will not be used
+    // and properties that will be recalculated for visualization
+    delete newNode._rev;
```
JackWilb, line 37: This is an Arango field and can likely be removed on import. Thoughts @waxlamp

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+
+    // add new node to node list
+    newNodes.push(newNode);
+
+    // print the new node and its attribute
+    // console.log("NEW NODE");
+    // console.log(newNode)
+  });
+  // create a new supernode and a new super node list
+  const superNodes: any[] = [
```
JackWilb, line 49: We could either make this type `Node[]` or we could add a new type, `SuperNode`, which would extend `Node` with the additional children property.

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+      ORIGIN_STATE: 'California',
+      _key: 'CA',
+      id: 'supernodes/CA',
+    },
+  ];
+
+  // update the index attribute of a superNode
+  superNodes.forEach((superNode, index) => {
+    superNode.index = index;
+  });
```
JackWilb, line 61: Do we need to specify the index? The Objects never move in the Array

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+
+  // print information about the super node list
+  // console.log("THE SUPER NODE LIST");
+  // console.log(superNodes);
```
JackWilb, line 65: Please remove

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+
+    // add new attributes for the new links
+    // id - attribute for visualization and determining the neighbors for the network
+    newLink.index = `edges/${index}`;
+
+    newLinks.push(newLink);
+
+    // print information about the new link
+    // console.log("NEW LINK");
+    // console.log(newLink);
```
JackWilb, line 99: Please remove

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+  // make a new list of edges for the supergraph network
+  const newLinks: any = [];
+  edges.forEach((link, index) => {
+    const newLink = {
+      ...link,
+    };
+
+    // remove attributes that will be re-calculated for the links
+    // and attributes that are not used for aggregation/visualization
+    delete newLink._rev;
```
JackWilb, line 89: This could also be removed on data loading, instead of here. See above

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+    newLinks.push(newLink);
+
+    // print information about the new link
+    // console.log("NEW LINK");
+    // console.log(newLink);
+  });
+
+  // // print information about the new links
+  // console.log('NEW LINKS');
+  // console.log(newLinks);
```
JackWilb, line 104: Please remove

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+        }
+      });
+    });
+  });
+
+  // print information about the new links and new nodes
+  // console.log('NEW NODES');
+  // console.log(newNodes);
+  // console.log('NEW LINKS');
+  // console.log(newLinks);
```
JackWilb, line 132: Please remove

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+
+  // combine the superNodes with the new nodes before updating all the neighbors
+  const combinedNodes = superNodes.concat(newNodes);
+
+  // print information about combined nodes
+  // console.log("THE COMBINED NODES");
+  // console.log(combinedNodes);
```
JackWilb, line 139: Please remove

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+
+  // construct the neighbors for the nodes
+  const neighborNodes = defineSuperNeighbors(combinedNodes, newLinks);
+
+  // print information about the neighbor nodes
+  // console.log('THE NEIGHBOR NODES');
+  // console.log(neighborNodes);
```
JackWilb, line 146: Please remove

src/components/Controls.vue
```diff
@@ -104,6 +252,19 @@ export default Vue.extend({
 
       legendSVG.select('.legendLinear').call(legendLinear);
     },
+    clickButton(this: any) {
+      // print that the button was clicked
+      // console.log('clicked the aggregate button');
+
+      // call the super graph function and print stuff
+      // const superNetwork = superGraph(this.network.nodes, this.network.links);
+      // // console.log(superNetwork);
```
JackWilb, line 261: Please remove

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+    delete newNode._rev;
+    delete newNode.index;
+    delete newNode.neighbors;
+
+    // add new node to node list
+    newNodes.push(newNode);
+
+    // print the new node and its attribute
+    // console.log("NEW NODE");
+    // console.log(newNode)
```
JackWilb, line 46: Please remove

src/components/Controls.vue
```diff
@@ -6,8 +6,156 @@ import { format } from 'd3-format';
...
+
+  // construct the neighbors for the nodes
+  const neighborNodes = defineSuperNeighbors(combinedNodes, newLinks);
+
+  // print information about the neighbor nodes
+  // console.log('THE NEIGHBOR NODES');
+  // console.log(neighborNodes);
+
+  // remove all the nodes who do not have any neighbors
+  const finalNodes = neighborNodes.filter((node) => node.neighbors.length != 0);
```
JackWilb, line 149: This looks like it could be a bug. There could be nodes that have no neighbors before and after manipulation

JackWilb (changes requested):
I noticed lots of dead code that should be removed, which made reviewing the logic a little tricky. I pointed out a couple of logic issues, but once you've cleaned them up, I'd like to take another pass over it.

src/components/AdjMatrix/AdjMatrix.vue
```diff
@@ -144,11 +147,9 @@ export default Vue.extend({
         window.innerHeight ||
         document.documentElement.clientHeight ||
         document.body.clientHeight;
-
       // Set dimensions of the node link
-      this.visDimensions.width = this.browser.width * 0.75;
-      this.visDimensions.height = this.browser.height - 24;
-
+      // this.visDimensions.width = this.browser.width * 0.75;
+      // this.visDimensions.height = this.browser.height - 24;
```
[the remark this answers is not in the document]
waxlamp: Agreed.

src/components/Controls.vue
waxlamp: I agree. We could also just ignore it.

src/lib/multinet.ts
mozartfish: This function needs to exist because if the _defineNeighbors d._id is changed to d.id, the visualization breaks. The reason is that when the data is originally uploaded by the user the id is represented with d._id. By the time the super node aggregation occurs, the id has already been converted from d._id to d.id using the data processing functions that are executed before the visualization is first rendered.

src/components/Controls.vue
waxlamp: I think `Node[]` is more appropriate at the moment. We'll have to think about whether we want to canonize supernodes as a specific type.

src/components/Controls.vue
waxlamp: This depends very much on why we need the index present. If it's for rendering the matrix, we need to figure out how to make sure the logic of computing the index is encapsulated to the rendering process.

Thoughts, @JackWilb and @mozartfish?

src/components/Controls.vue
```diff
@@ -6,8 +6,120 @@ import { format } from 'd3-format';
...
+
+  // update the parent field of the node if it has a super node with the super node id
+  // update the super node origin list with the child node id
+  newNodes.forEach((node) => {
+    if (node.ORIGIN_STATE === 'California') {
+      const superNode = superNodes.find(
+        (superNode) => superNode.ORIGIN_STATE === 'California',
+      );
+      node.parent = superNode.id;
+      superNode.ORIGIN.push(node.id);
```
waxlamp, line 61: I don't think this is correct, mainly because the data type will now vary based on whether this is a node or a supernode. Maybe that's unavoidable. But it makes thinking about the types involved a lot harder, so that makes me want to re-evaluate how we do this.

src/components/Controls.vue
```diff
@@ -6,8 +6,120 @@ import { format } from 'd3-format';
...
+      ...link,
+    };
+
+    // remove attributes that will be re-calculated for the links
+    // and attributes that are not used for aggregation/visualization
+    delete newLink._rev;
+
+    // add new attributes for the new links
+    // id - attribute for visualization and determining the neighbors for the network
+    newLink.index = `edges/${index}`;
```
waxlamp, line 79: The comment seems to disagree with the line of code. This looks more like an ID value to me, but it's drawing from an index value. I think we need a better sense for what the semantics of all these attributes is supposed to be.

src/components/Controls.vue
```diff
@@ -6,8 +6,120 @@ import { format } from 'd3-format';
...
+          link.source = link._from;
+        }
+        if (linkTo === origin) {
+          const newLinkTo = superNode.id;
+          link._to = newLinkTo;
+          link.target = link._to;
+        }
+      });
+    });
+  });
```
waxlamp, line 104: This is still very confusing.

I think it's probably fine for now, but we need to rearchitect how data flows through the system. I think the component should track the original data from Multinet, and use a computed prop to compute what is needed for D3. Then, this function wouldn't have to transform data from Multinet format to D3 format here and now; it could instead do the Multinet-to-Multinet transform, and let the component handle the conversion to D3.

Does this make sense @JackWilb?

src/components/Controls.vue
waxlamp: The real thing to do is to go through the supernodes' children lists, and remove those nodes from `neighborNodes` (I think).

src/components/Controls.vue
```diff
@@ -104,6 +216,12 @@ export default Vue.extend({
 
       legendSVG.select('.legendLinear').call(legendLinear);
     },
+    clickButton(this: any) {
```
waxlamp, line 219: This should really be called "aggregateCalifornia" or something like that.

src/components/Controls.vue
```diff
@@ -104,6 +216,12 @@ export default Vue.extend({
...
+      // call the supergraph function to generate the new nodes and the new edges for the super graph
+      // assign to a variable that can update the network which updates the visualization
+      const superNetwork = superGraph(this.network.nodes, this.network.links);
+      this.network = superNetwork;
```
waxlamp, line 223: A general note about comments: you shouldn't self-describe the "obvious" parts of what the code is doing. In this case, you don't need to say you're calling a function and assigning to a variable. I'd simplify things like this:

```suggestion
      // Compute a new graph based on aggregating California airports into a supernode.
      this.network = superGraph(this.network.nodes, this.network.links);
```

src/components/Controls.vue
```diff
@@ -178,6 +296,11 @@ export default Vue.extend({
             </v-card-subtitle>
           </v-card-text>
 
+          <!--button for aggregating by california -->
```
waxlamp, line 299: Another example: note that this comment doesn't add anything that's not self-evident from the code.

```suggestion
```

src/lib/multinet.ts
waxlamp: I doubt this was from the linter. I'd leave the line here (in the interest of both using whitespace for readability, and not making code changes unrelated to the PR).

src/lib/multinet.ts
```diff
@@ -32,19 +32,15 @@ async function _downloadAllRows(
...
   return output;
 }
 
-function _renameLinkVars(links: any[]): Link[] {
+export function _renameLinkVars(links: any[]): Link[] {
```
waxlamp, line 38: If this is to be a public function, then remove the underscore.

src/lib/multinet.ts
```diff
@@ -66,6 +62,16 @@ function _defineNeighbors(nodes: any[], links: any[]) {
   return nodes;
 }
 
+// Function that constructs the neighbors for a node in a super network
+export function defineSuperNeighbors(nodes: any[], links: any[]) {
+  nodes.map((d: { neighbors: string[] }) => (d.neighbors = []));
+  for (const link of links) {
```
waxlamp, line 68: We should prefer `Array.forEach()` to `for..of` loops.

src/lib/multinet.ts
```diff
@@ -66,6 +62,16 @@ function _defineNeighbors(nodes: any[], links: any[]) {
...
+    nodes.filter((d: Node) => d.id === link._from)[0].neighbors.push(link._to);
+    nodes.filter((d: Node) => d.id === link._to)[0].neighbors.push(link._from);
+  }
+  return nodes;
+}
```
waxlamp, line 73: I'm having some trouble understanding what this function does. The twin filter statements seem strange to me: they perform a filter on the full nodes array, then only operate on the first detected element. It seems inefficient, and the tight functional style obscures what this is trying to do.

waxlamp (changes requested):
Looking good, @mozartfish. Check out Jack's and my suggestions, and involve us if anything is unclear or if you want to discuss any of our feedback.

src/lib/multinet.ts
mozartfish: @JackWilb @waxlamp from my understanding of functional programming and what this function is doing (based on the original defineNeighbors) this function is identifying and creating the array of neighbor nodes for each node in the network. The reason why I had to create this function was that the data format changed between uploading the data and when the vis was first created before aggregation (the ._to, ._from). I think there might be cleaner way how to handle this since I found it slightly confusing to identify which parts needed to be kept for aggregation. Maybe we can clean up the defineNeighbors and defineSuperNeighbors in such a way where we don't have to duplicate code? I haven't found a solution for this though

src/components/Controls.vue
waxlamp: This was fixed in an unknown commit.

src/components/Controls.vue
waxlamp: Opened #118 to track this, because for now we actually need to retain this function.

src/lib/multinet.ts
```diff
@@ -32,37 +32,43 @@ async function _downloadAllRows(
...
+  });
+  return nodes;
+}
+
+// Function that constructs the neighbors for a node in a super network
+export function defineSuperNeighbors(nodes: any[], links: any[]) {
+  nodes.map((d: { neighbors: string[] }) => (d.neighbors = []));
+  links.forEach((link) => {
+    nodes.filter((d: Node) => d.id === link._from)[0].neighbors.push(link._to);
+    nodes.filter((d: Node) => d.id === link._to)[0].neighbors.push(link._from);
```
waxlamp, line 70: Use `.find()` instead of `.filter()` followed by `[0]`.

waxlamp: Opened #119 to track this.

src/components/Controls.vue
```diff
@@ -1,10 +1,177 @@
...
+  // make a new list of nodes
+  const newNodes: any[] = [];
+  nodes.forEach((node, index) => {
+    const newNode = {
+      ...node,
+      // add new attributes for the new nodes
+      // parent - attribute for identifying the supernode parent
+      // index - attribute for keeping track of the index for visualizing the network
+      parent: undefined,
+      index: index,
```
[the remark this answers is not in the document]
mozartfish: fixed this issue. originally I thought you were referring to the parent so I changed the logic so we didn't need the parent but I just figured out now that index is not being used as I thought it was so it makes sense to remove this variable https://github.com/multinet-app/multimatrix/pull/110/commits/1cf25f3c37b26758d00fe36a7e4b73d3e665c40b

mozartfish:
@waxlamp @JackWilb according to this comment https://github.com/multinet-app/multimatrix/pull/110#discussion_r510333544, I found that the index is note needed for supernodes for updating the visualization so I removed the code in this commit https://github.com/multinet-app/multimatrix/pull/110/commits/8494c9dd316f56f2d41b5e3cc2e44c9cd201b051

mozartfish:
@waxlamp ran the linter and fixed the issue in this comment https://github.com/multinet-app/multimatrix/pull/110#discussion_r510342785

mozartfish:
This comment has been resolved https://github.com/multinet-app/multimatrix/pull/110#discussion_r510333544

src/components/Controls.vue
mozartfish: resolved with this commit

src/components/Controls.vue
mozartfish: this issues has been resolved with commit https://github.com/multinet-app/multimatrix/pull/110/commits/0bf49564bff83f4af90ad6879c95f16998797c4a

src/components/Controls.vue
mozartfish: I commented on this issue recently

src/lib/multinet.ts
mozartfish: fixed the error with the linter by running yarn lint:fix

src/components/Controls.vue
mozartfish: I agree with Roni and I think this might happen when we get closer to finalizing the aggregation grouping functionality

src/components/Controls.vue
mozartfish: this has been resolved and the index has been removed

waxlamp (approved):

JackWilb (approved):

PR #116, opened 2020-10-29
Rename adjmatrix to MultiMatrix

JackWilb:


package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "vue-adjmatrix",
+  "name": "MultiMatrix",
```
waxlamp, line 2: This name will be what NPM uses to publish this as a package. If we were to ever publish this, we'd want to use the lowercase name (same as the GitHub repo).

```suggestion
  "name": "multimatrix",
```

package.json
```diff
@@ -1,8 +1,8 @@
...
   "version": "0.1.0",
   "private": true,
-  "description": "An adjacency matrix visualization built to fit in with multinet",
+  "description": "A first class MultiNet adjacency matrix visualization",
```
waxlamp, line 5: What do you mean by `first class` here? I would suggest:

```suggestion
  "description": "An adjacency matrix visualization for MultiNet",
```

waxlamp (changes requested):
Thanks for doing this, @JackWilb. I just have a couple of suggestions but everything else looks great.

waxlamp (dismissed):

waxlamp (approved):

PR #120, opened 2020-11-11
Attribute selection

mozartfish:
This branch attempts to generalize aggregation by grouping for different attributes of the flight dataset.

PR #121, opened 2020-11-11
Allow disabling and enabling of graffinity aggregation features

JackWilb:
Depends on #120.

Adds a boolean flag to enable/disable the superGraph creation, and some UI to manage the toggling. Currently, this code just disables the attribute column header selection, since @mozartfish told me that the "Aggregate California" button will be removed before he merges #120. This should disable any call to `superGraph()`, thus disabling the graffinity features

Leaving as a draft for now, until @mozartfish merges #120.

Closes https://github.com/orgs/multinet-app/projects/6#card-48781425

### DIFF
--- a/src/components/AdjMatrix/AdjMatrixMethods.ts
+++ b/src/components/AdjMatrix/AdjMatrixMethods.ts
@@ -144,7 +144,6 @@ export class View {
       .attr('x', (d: string, i: number) => (colWidth + this.colMargin) * i)
       .attr('width', colWidth)
       .on('click', (d: string) => {
-        console.log('clicked the text label');
         if (this.enableGraffinity) {
           this.network = superGraph(this.network.nodes, this.network.links, d);
           eventBus.$emit('updateNetwork', this.network);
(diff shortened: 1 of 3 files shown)


waxlamp (approved):
This is nice. I was thinking that a query argument to enable the graffinity features would be better (so that the app would just look like regular multimatrix unless you knew the secret query argument) but I think this is good too.

Later on, we'll either want to convert this to my query arg idea, or just have multimatrix organically gain all of the the experimental graffinity features.

Because #120 is still under review, I assume this PR may need to be rebased or otherwise updated before final merge.

Thanks @JackWilb!

src/components/AdjMatrix/AdjMatrixMethods.ts
```diff
@@ -141,8 +144,11 @@ export class View {
       .attr('x', (d: string, i: number) => (colWidth + this.colMargin) * i)
       .attr('width', colWidth)
       .on('click', (d: string) => {
-        this.network = superGraph(this.network.nodes, this.network.links, d);
-        eventBus.$emit('updateNetwork', this.network);
+        console.log('clicked the text label');
```
waxlamp, line 147:
```suggestion
```

waxlamp (changes requested):
There's a stray `console.log()` that we should get rid of, but otherwise this looks great.

waxlamp (approved):

PR #133, opened 2020-11-30
Merge graffinity development branch into master

### LINKED ISSUE #129: Implementing Sorting for Supergraph Nodes
Sorting is a data wrangling operation that was discussed in Origraph, Graffinity, and the State of the Art Report on Multivariate Networks but hasn't been implemented yet for supernodes. This is a possible feature to add to MultiNet for supernode aggregation and addresses this comment https://github.com/multinet-app/multimatrix/pull/120#discussion_r528043676_

JackWilb:
Since we've added a way to enable/disable experimental graffinity features, we're now at a good point to merge the `graffinity` branch into master. 

This gives us a few benefits:
1. Deploy previews and CI for new code
2. We don't need to manage an additional branch
3. Changes will all be located in one place (instead of across multiple branches). Thus critical changes to master will always be available, and we won't need to keep rebasing `graffinity` onto `master`

A couple of things we should address:
- [x] It looks like Jared's inital commits that modified the sidebar have been reverted (#134)
- [x] Sorting using the attributes no longer works (in experimental mode or not) (Closes #129)
- [x] Merge conflicts


waxlamp (approved):
Whew! Thanks for shepherding this through, Jack!

PR #135, opened 2020-12-04
Fix any type issue for the functions in aggregation.ts Addresses Issue#<123> and MapNetworkNodes

mozartfish:
This PR fixes the issues with the any type in aggregation.ts and makes it easier to know what the types of objects are when using these specific functions for aggregation and adds a new function for keeping track of nodes and ids in a network

### DIFF
--- a/src/lib/aggregation.ts
+++ b/src/lib/aggregation.ts
@@ -4,7 +4,7 @@
 // in the network
 import { defineSuperNeighbors } from '@/lib/multinet';
 import { Link, Node } from '@/types';
-export function superGraph(nodes: any[], edges: any[], attribute: string) {
+export function superGraph(nodes: Node[], edges: Link[], attribute: string) {
   // de-construct nodes into their original components and
   // make a new list of nodes
   const newNodes: Node[] = [];
@@ -15,7 +15,7 @@ export function superGraph(nodes: any[], edges: any[], attribute: string) {
 
     // remove the properties that will not be used
     // and properties that will be recalculated for visualization
-    delete newNode.neighbors;
+    newNode.neighbors = [];
 
     // add new node to node list
     newNodes.push(newNode);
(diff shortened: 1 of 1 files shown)


src/lib/aggregation.ts
```diff
@@ -15,7 +15,7 @@ export function superGraph(nodes: any[], edges: any[], attribute: string) {
 
     // remove the properties that will not be used
     // and properties that will be recalculated for visualization
-    delete newNode.neighbors;
+    newNode.neighbors = [];
```
waxlamp, line 18: Why this change?

src/lib/aggregation.ts
```diff
@@ -112,3 +112,13 @@ export function superGraph(nodes: any[], edges: any[], attribute: string) {
...
+
+// function that maps node names to nodes
+export function MapNetworkNodes(nodes: Node[]) {
+  const nodeMap = new Map<string, Node>();
+  nodes.forEach((node: Node) => {
+    nodeMap.set(node.id, node);
+  });
+
+  return nodeMap;
+}
```
waxlamp, line 124: This doesn't seem like it belongs in this PR.

waxlamp:
@mozartfish, can you edit the PR description and include the phrase `Closes #<issue number>` so that the appropriate issue or issues are closed when this is merged?

src/lib/aggregation.ts
mozartfish: [
Screenshot from 2020-12-04 11-06-44
]

mozartfish: there was so many comments about this issue on github so I tried some stuff and got this error to go away when I set the neighbors to empty

mozartfish: according to this post https://stackoverflow.com/questions/63702057/what-is-the-logic-behind-error-the-operand-of-a-delete-operator-must-be-opti, when changing the  type of nodes from any to node, I violated the interface contract for node since it requires a field called neighbors. What I did with this change to maintain the contract of the interface is set an empty list of neighbors so that the neighbors field would not be undefined and there would not be a contact violation.

mozartfish: https://stackoverflow.com/questions/63702057/what-is-the-logic-behind-error-the-operand-of-a-delete-operator-must-be-opti

PR #138, opened 2020-12-11
Reimplement sidebar with better styling

### LINKED ISSUE #134: Rework Jared's sidebar from #82
In #82, Jared created some UI for a 1 hop network that was more inline with the multinet.app theme. We chose to remove this work and keep our current style for the time being, because we didn't need the functionality. However, we should rework the sidebar into something that we can use to ensure consistent theming across the app

JackWilb:
Closes #134 

This adds back the sidebar that Jared created with the controls from the main app. I had to set a specific width on the logo image to match the main client, but I think it looks okay.

@waxlamp Should we add the icon for logins here? We could also add the about dialog from the main client, too. Let me know what you think. I also added a section for future widgets, should I remove that from the PR?

@jtomeck I have a couple of questions for you (that are also in the TODO section).
  * Something is off about the padding, but I'm not certain exactly what it is. Would you mind taking a pass on this code and paying special attention to this?
  * The font here is different from the main app, should we change the font here or on the main app for consistency?

TODO:
- [x] Fix the matrix legend size, it's too big as it is. It might also some alignment (@JackWilb)
- [x] Decide whether we should remove the widget section (@waxlamp)
- [x] Fix general padding through the controls (@jtomeck)
- [x] Decide whether we still want to use this custom font. If we do it needs to be propagated to the main client, else we could revert it to use the fonts from the main client (@jtomeck)
- [x] Alex requested that the menu list items be "more compact" since we'll probably want more controls than we currently have (@JackWilb make an issue) (#154)


jtomeck:
@JackWilb - I've made a commit that helps a lot, although I don't have any data loaded so let me know if there's more that you think is needed.

I remember looking at this font a while back. Did we ever decide to commit to the new font change? I'm leaning towards keeping the old because I think it's causing alignment issues in components like buttons.

JackWilb:
> @JackWilb - I've made a commit that helps a lot, although I don't have any data loaded so let me know if there's more that you think is needed.

Looks great, the styling should not be affected by data loading (minus the "matrix legend" piece, which I'm planning to address). Thanks 😄 

> I remember looking at this font a while back. Did we ever decide to commit to the new font change? I'm leaning towards keeping the old because I think it's causing alignment issues in components like buttons.

@jtomeck I believe the original font we were using was Avenir, and that's not a free font. The license cost looks to be [300 dollars](https://www.fonts.com/font/linotype/avenir/packages) for the package. We should choose a font that doesn't give us these alignment issues and then roll out the change for the whole app ecosystem

jtomeck:
I really love the way this font looks. It's seemingly just the text alignment inside buttons that is an issue. It's a couple pixels below vertical centered. It's not a huge issue, but something I'd like to address when there's more funding (I'm not sure we'll want to spend too much time on this right now, but @waxlamp can correct me if I'm wrong). So I guess we just roll with this font for now and I can create an issue for button text alignment?

waxlamp:
> I really love the way this font looks. It's seemingly just the text alignment inside buttons that is an issue. It's a couple pixels below vertical centered. It's not a huge issue, but something I'd like to address when there's more funding (I'm not sure we'll want to spend too much time on this right now, but @waxlamp can correct me if I'm wrong). So I guess we just roll with this font for now and I can create an issue for button text alignment?

It sounds like you have a concrete thing you can fix with some targeted time--I'd say file a quick issue, then revisit in January. If it's critical to fix it sooner, go ahead and do it, just let me know how much time you think it will take before you start. (This doesn't sound critical; correct me if I'm wrong.)

jtomeck:
> It sounds like you have a concrete thing you can fix with some targeted time--I'd say file a quick issue, then revisit in January. If it's critical to fix it sooner, go ahead and do it, just let me know how much time you think it will take before you start. (This doesn't sound critical; correct me if I'm wrong.)

Yeah it's not critical. To be honest, let me know if you guys are even noticing it. It's very slight. Here's two screenshots. One with Roboto and one with Blinker.

Screen Shot 2020-12-14 at 1 06 42 PM
Screen Shot 2020-12-14 at 1 08 17 PM

waxlamp:
> Yeah it's not critical. To be honest, let me know if you guys are even noticing it. It's very slight. Here's two screenshots. One with Roboto and one with Blinker.
> 
> Screen Shot 2020-12-14 at 1 06 42 PM Screen Shot 2020-12-14 at 1 08 17 PM

I wouldn't notice anything off unless you had mentioned it, and even with you mentioning it, I'm not sure I'm seeing it. The one of the left seems to be maybe a bit lower than center (did I get it right?).

FWIW, I really like the font in the left screencap. I feel it fits our project better than the one on the right.

jtomeck:
Yeah you got it right, and I agree. I want to use this font, even with the vertical alignment issue. I posted a hacky fix in my comment above if you guys think it's worth.

waxlamp:
> Yeah you got it right, and I agree. I want to use this font, even with the vertical alignment issue. I posted a hacky fix in my comment above if you guys think it's worth.

Sorry, I missed the suggested fix in my first go-round. If you don't think that fix presents too much technical debt, then I'm ok with it. Need @JackWilb's opinion.

JackWilb:
> > Yeah you got it right, and I agree. I want to use this font, even with the vertical alignment issue. I posted a hacky fix in my comment above if you guys think it's worth.
> 
> Sorry, I missed the suggested fix in my first go-round. If you don't think that fix presents too much technical debt, then I'm ok with it. Need @JackWilb's opinion.

Seems reasonable for now. We'll have to keep an eye on it and make sure that it's the only place where we see the alignment issue. If we start seeing it crop up in more places, the better solution would just be to choose a different font that is aligned properly

waxlamp:
> > > Yeah you got it right, and I agree. I want to use this font, even with the vertical alignment issue. I posted a hacky fix in my comment above if you guys think it's worth.
> > 
> > 
> > Sorry, I missed the suggested fix in my first go-round. If you don't think that fix presents too much technical debt, then I'm ok with it. Need @JackWilb's opinion.
> 
> Seems reasonable for now. We'll have to keep an eye on it and make sure that it's the only place where we see the alignment issue. If we start seeing it crop up in more places, the better solution would just be to choose a different font that is aligned properly.

I think changing the font just to fix a small alignment problem is overkill but it could be a solution.

> 
> I'll make some issues in the other repo to track this.

Which repo?

JackWilb:
> I think changing the font just to fix a small alignment problem is overkill but it could be a solution.

Not if the alignment issue is in the font, which I think might be the issue. We should just keep an eye on it

> > I'll make some issues in the other repo to track this.
> 
> Which repo?

Sorry, I meant for tracking the font migration for each of the repos, so multinet-client and multi-link. We should have a consistent font across the whole ecosystem.

Do you think we need an issue here to track the potential alignment issue, too?

JackWilb:
Met with Alex and he request that the menu be "more compact"

waxlamp:
> @waxlamp Should we add the icon for logins here?

I'm not sure what to do about a login icon. I am thinking it doesn't belong in this PR in any case.

How would it work? Can we redirect folks to the OAuth login sequence for Multinet client but have them redirected to this app? Sounds like a question @AlmightyYakob could weigh in on. @JackWilb, since this will take some thinking and discussion, I would file an issue and defer discussions and decisions to there.

> We could also add the about dialog from the main client, too. Let me know what you think.

An about dialog sounds like a good idea. But again, I'd defer it to future work.

> I also added a section for future widgets, should I remove that from the PR?

Yes I think so. Since this is going to production, an empty "widgets" section just looks incomplete. Is there some concrete plan for what we want to put in here?

JackWilb:
> > @waxlamp Should we add the icon for logins here?
> 
> I'm not sure what to do about a login icon. I am thinking it doesn't belong in this PR in any case.
> 
> How would it work? Can we redirect folks to the OAuth login sequence for Multinet client but have them redirected to this app? Sounds like a question @AlmightyYakob could weigh in on. @JackWilb, since this will take some thinking and discussion, I would file an issue and defer discussions and decisions to there.

Opened #155 

> > We could also add the about dialog from the main client, too. Let me know what you think.
>
> An about dialog sounds like a good idea. But again, I'd defer it to future work.

Opened #156

> > I also added a section for future widgets, should I remove that from the PR?
> 
> Yes I think so. Since this is going to production, an empty "widgets" section just looks incomplete. Is there some concrete plan for what we want to put in here?

No explicit plans. I replaced it with "Color Scale Legend" and moved the legend out of the controls section. I think that makes more sense and if we have widgets, we can add it later.

waxlamp (approved):

Issue #141, opened 2020-12-16
Remove the class logic from the app


JackWilb:
The class logic has become a major bottleneck in our future work on the matrix repo and it needs to be removed. The initial plan was to make this change as one big PR, but I think I can do it as multiple smaller ones.

PR #142, opened 2020-12-16
Move processData and some helpers to MultiLink.vue

JackWilb:
Relates to #141 

This is the first of many changes to remove the class logic and moves the `processData` function out of the class. To do this, I needed to set up a few `data` variables in the vue component and move the logic. Since some of the variables (`idMap` and `matrix`) need to be regenerated when the network changes, I separated some of the logic out into functions. That lets me call them in multiple places, e.g. on mount and when the matrix updates for graffinity aggregation.

I also tried to simplify some logic where I saw fit by moving the class variable instantiation out of the constructor, and removing matrix node length.

### DIFF
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -137,7 +137,6 @@ export default Vue.extend({
     this.view = new View(
       this.network,
       this.visualizedAttributes,
-      this.matrixNodeLength,
       this.cellSize,
       this.visMargins,
       this.enableGraffinity,
@@ -188,7 +187,6 @@ export default Vue.extend({
       this.view = new View(
         this.network,
         this.visualizedAttributes,
-        this.matrixNodeLength,
         this.cellSize,
         this.visMargins,
         this.enableGraffinity,
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -3,7 +3,7 @@ import * as d3 from 'd3';
 import Vue, { PropType } from 'vue';
 
 import { View } from '@/components/MultiMatrix/MultiMatrixMethods';
-import { Dimensions, Network } from '@/types';
+import { Cell, Dimensions, Link, Network, Node } from '@/types';
 
 export default Vue.extend({
   props: {
@@ -28,10 +28,13 @@ export default Vue.extend({
   data(): {
     browser: Dimensions;
     visMargins: any;
-    matrix: any;
+    matrixSVG: any;
     attributes: any;
     view: View | undefined;
     cellSize: number;
+    idMap: { [key: string]: number };
+    maxNumConnections: number;
+    matrix: Cell[][];
   } {
     return {
       browser: {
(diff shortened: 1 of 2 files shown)


mozartfish:
does this code right here https://github.com/multinet-app/multimatrix/pull/142/commits/d55b726b1877c39aec53e5ea2b90af77d89cfb0b#diff-42cea2605481e7c9660c0b91eed182e5559253fc5306341b206d586d15fdf8b3R194-R198 fix the problem with two different ids? By this I mean that instead of having _id and id like earlier, the id attribute is now access with node.id.

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -171,12 +186,52 @@ export default Vue.extend({
...
       );
     },
+
+    generateIdMap() {
+      this.idMap = {};
+      this.network.nodes.forEach((node: Node, index: number) => {
+        node.index = index;
+        this.idMap[node.id] = index;
+      });
+    },
```
waxlamp, line 203: It seems like this could be a computed prop. It's a pure function of `this.network.nodes` (aside from the "reindexing" that's happening here, which on its own seems like it could go into a computed prop as well) and it's only updated in a watcher on `this.network`.

If we're able to make this suggested change, I believe it would make the codebase cleaner and simpler to understand. If you agree that it's a good idea you could either do it now or (more likely) file an issue to tackle it later.

I'll approve this for now and you can decide how to proceed.

waxlamp (approved):

waxlamp:
> does this code right here [d55b726#diff-42cea2605481e7c9660c0b91eed182e5559253fc5306341b206d586d15fdf8b3R194-R198](https://github.com/multinet-app/multimatrix/commit/d55b726b1877c39aec53e5ea2b90af77d89cfb0b#diff-42cea2605481e7c9660c0b91eed182e5559253fc5306341b206d586d15fdf8b3R194-R198) fix the problem with two different ids? By this I mean that instead of having _id and id like earlier, the id attribute is now access with node.id.

Since we're in a PR here, you can comment directly on those lines of the diff, instead of including a link. It might make discussion a little easier.

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: Opened https://github.com/multinet-app/multimatrix/issues/143

mozartfish (approved):

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -171,12 +186,52 @@ export default Vue.extend({
...
+            z: 0,
+          };
+        });
+      });
+
+      // Count occurrences of links and store it in the matrix
+      this.network.links.forEach((link: Link) => {
+        this.matrix[this.idMap[link.source]][this.idMap[link.target]].z += 1;
+        this.matrix[this.idMap[link.target]][this.idMap[link.source]].z += 1;
+      });
```
gotdairyya, line 224: Flagging this for a few reasons: 

1. This bit of code I believe is responsible for the matrix double-counting directed edges (at least part of it). Though reading through this logic it doesn't seem like it should double count..
2. We could use the neighbors attribute to compute the number of neighbors
3. link.target and link.source are relics of d3-isms. Do we want to switch to _to and _from (more native to Arango)?

JackWilb: 1. This just means that the edges are not directional and it shouldn't lead to over counting. We'll need to add a toggle for directional edges (opened #145)
2. We can't. The neighbors just track if there is at least one connection. We instead need to count the total number of connections.
3. I agree (opened #144)

gotdairyya (approved):
aside from keeping track of double-counting/visualizing directed edges, all looks good!

PR #146, opened 2020-12-17
Migrate initializeAttributes to MultiMatrix

JackWilb:
Another step towards #141 

This moves the initialize attributes function to MultiMatrix.vue.

I had to duplicate some logic: `showTooltip`, `hideTooltip`, `isCell`, `capitalizeFirstLetter`, `hoverNode`, `unHoverNode`. The best strategy (that I could think of) was to duplicate the logic, since the code is used in both `initializeAtrributes` and `initializeEdges`. I'll remove the duplicated code when the `initializeEdges` is moved.

I also commented out the code for on click for the attribute rows to reduce the amount of duplicated code. I'll also re-enable that once I move `initializeEdges`.

This introduces a new bug that I'll have to track down. Sorting a aggregated network causes there to be a bunch of blank rows. I would guess that it's to do with how the network is being updated. I'll open an issue to track this before I merge and address this in my follow up PRs.

It might be time to target this to a longer running branch that tracks the rest of these removals (so we don't lose sight of fixing the bugs before we merge)

TODO:
- [x] Sorting aggregated network causes there to be some blank rows (open issue @JackWilb)

### DIFF
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -154,6 +154,8 @@ export default Vue.extend({
         `translate(${this.visMargins.left},${this.visMargins.top})`,
       );
 
+    this.initializeAttributes();
+
     // Define the View
     this.view = new View(
       this.network,
@@ -216,6 +218,8 @@ export default Vue.extend({
           `translate(${this.visMargins.left},${this.visMargins.top})`,
         );
 
+      this.initializeAttributes();
+
       this.view = new View(
         this.network,
         this.visualizedAttributes,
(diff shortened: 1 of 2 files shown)


src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -29,12 +29,16 @@ export default Vue.extend({
...
+    attributesSVG: any;
     view: View | undefined;
     cellSize: number;
     idMap: { [key: string]: number };
     maxNumConnections: number;
     matrix: Cell[][];
+    attributes: any;
+    attributeRows: any;
+    columnHeaders: any;
+    edges: any;
```
waxlamp, line 41: Is it possible to include types here?

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -232,6 +275,146 @@ export default Vue.extend({
         });
       });
     },
+
+    initializeAttributes(): void {
+      const attributeWidth = 1000; // Just has to be larger than the attributes panel (so that we render to the edge)
```
waxlamp, line 280: Generally, stylewise, I think it's better not to have these "trailing" comments (in this case, I'd just move the comment to its own line above).

src/components/MultiMatrix/MultiMatrixMethods.ts
```diff
@@ -86,6 +80,11 @@ export class View {
     enableGraffinity: boolean,
     matrix: Cell[][],
     maxNumConnections: number,
+    orderingScale: ScaleBand<number>,
+    columnHeaders: any,
+    edges: any,
+    attributes: any,
+    attributeRows: any,
```
waxlamp, line 87: Why is stuff getting added to the class? I'd have expected deletions instead?

src/components/MultiMatrix/MultiMatrixMethods.ts
```diff
@@ -376,11 +366,51 @@ export class View {
         this.hideToolTip();
         this.unHoverEdge(d);
       })
-      .on('click', this.selectElement)
+      .on('click', (d: Cell) => this.selectElement(d))
```
waxlamp, line 369: Why make this change?

waxlamp (approved):
Looking good.

As for creating a long-running branch to collect these, that's probably a good idea, but keep in mind the lesson we learned before: if master diverges much from that branch, there's pain involved in the merge step. So even if it's a long-running branch, let's try to make it a short-lived long-running branch, as it were.

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: It's possible, probably, but it's D3 types. I didn't want to get bogged down dealing with them now, while we make the switch. If it's okay with you, I'd like to open an issue to revisit this once the rest of the code is moved.

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: Fixed in 8802842

src/components/MultiMatrix/MultiMatrixMethods.ts
JackWilb: This is just to maintain compatibility from the vue component to the class. Since I removed the definitions from inside methods.ts, I had to sync them up by passing the values into the constructor. Lots of these should be removed in the next PR, when I migrate the `initializeEdges` function

src/components/MultiMatrix/MultiMatrixMethods.ts
JackWilb: This was a bug I noticed -- cell clicking wasn't working. I can back this change out of this PR, if you'd like and open a new one for it

waxlamp: No it's fine, since the codebase is in a kind of upheaval state right now anyway. I'm just puzzled that this would fix anything! Maybe the second parameter that d3 passes to the callback was causing trouble in `this.selectElement()`.

src/components/MultiMatrix/MultiMatrixMethods.ts
waxlamp: Got it.

src/components/MultiMatrix/MultiMatrix.vue
waxlamp: Yeah, file an issue and if d3 types remain as garbage as they are, we can just add a comment explaining that we're opting out of typing here just because of that.

waxlamp (approved):

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: Opened #147

gotdairyya (commented):
looks good!

PR #149, opened 2020-12-17
Move initializeEdges to MultiMatrix

JackWilb:
Another step to #141 

This moves `initializeEdges` to MultiMatrix.vue and all the supporting logic. This removes some of the duplicated logic from #146. To complete this move as painlessly as possible, I tried to move all the helper functions, move the main function, and then remove all the helper logic from the class.

The remaining logic in Methods.ts is just related to updating attributes, which I'll address in my next PR. That should fully remove the class.


JackWilb:
Had to do some force pushing to trigger GA and netlify build (to show that the code does still build)

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -219,21 +270,21 @@ export default Vue.extend({
           `translate(${this.visMargins.left},${this.visMargins.top})`,
         );
 
+      this.provenance = this.setUpProvenance();
```
gotdairyya, line 273: line 208 initializes this.provenance already, does it need to be reinitialized?

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -276,6 +327,436 @@ export default Vue.extend({
...
+        .attr('transform', `translate(0, 0)`);
+
+      this.edgeRows
+        .transition()
+        .duration(1100)
+        .attr('transform', (d: Node, i: number) => {
+          return `translate(0,${this.orderingScale(i)})`;
+        });
+
+      this.drawGridLines();
```
gotdairyya, line 392: we had discussed that you would add the exit remove logic for the matrix elements, is the plan to do that later?

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: I'm copying the functionality from the class. When it was torn down and rebuilt, the provenance would get reset. This code does that resetting. I'm planning to make many improvements to how we use provenance, including a change where we can keep the provenance state as we update what's rendering, but that's outside the scope of this PR.

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: Yeah, these PRs are just to move code around. This is an enhancement that will come later

JackWilb: Opened #150 to track it

waxlamp (approved):

PR #157, opened 2020-12-18
Fix capitalization and chip size in the control panel

JackWilb:
A couple small fixes. The title says it all 😄

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -158,7 +158,8 @@ export default Vue.extend({
               chips
               dense
               deletable-chips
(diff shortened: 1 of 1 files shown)


waxlamp (approved):
A note for future PRs: when possible please provide a link that demonstrates the branch on live data, which helps to review small style PRs like this one. In this case, clicking on the Netlify preview link brings up an "empty" app since we don't have an easy way to know which datasets are available in the testing server.

Even when a live preview is provided, before and after screencaps might even be more useful, since they show the change in context. (In this case I "reviewed" by finding out what small chips look like at the Vuetify documentation page for v-chip.)

PR #158, opened 2020-12-20
Expand Supernodes for matrix visualization

mozartfish:
This PR introduces a new feature that allows users to expand/retract the supernodes of an aggregated matrix visualization and view the connections between children nodes and supernodes


src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -69,6 +69,7 @@ export default Vue.extend({
     sortKey: string;
     colMargin: number;
     attributeScales: { [key: string]: any };
+    clickMap: any; // variable for keeping track of whether a label has been clicked or not
```
gotdairyya, line 72: would it work to have the type here as?

`{ [key: string]: Boolean }`

PR #160, opened 2020-12-21
Add toggle to enable edge directionality

### LINKED ISSUE #145: Add toggle for edge directionality
Currently all edges are treated as non-driectional, and thus our matricies are all symmetric. We should add a toggle to fix that.

JackWilb:
Closes #145 

Prior to this PR, edges were being counted in both directions. This is fine when that information is not directional (like a relationship in the les mis dataset), but when the edges encode a specific directional relationship we want to see that. This PR addresses that issue.

I added a toggle for edge directionality and updated the code to only count in one direction if directional edges is on. This code required a bit of duplicated logic for watching the `directional` prop and updating the matrix. Once we have a better strategy for handling updates to props (enter, exit, merge, etc.) we can refactor this so that we don't need to tear the matrix down and rebuild it.

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -24,6 +24,7 @@ export default Vue.extend({
     selectNeighbors: boolean;
     showGridLines: boolean;
     enableGraffinity: boolean;
+    directional: boolean;
     visualizedAttributes: string[];
   } {
     return {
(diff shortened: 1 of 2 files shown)


JackWilb:
Force pushed to fix a formatting issue from prettier

mozartfish (approved):
Looks good!

gotdairyya (approved):

waxlamp (approved):

PR #162, opened 2020-12-21
Prefer _from and _to to source and target

### LINKED ISSUE #144: Prefer using _to and _from instead of source and target
link.target and link.source are relics of d3-isms. Do we want to switch to _to and _from (more native to Arango)?

_Originally posted by @gotdairyya in https://github.com/multinet-app/multimatrix/pull/142#discussion_r544641193_

JackWilb:
Depends on #160 (Since the logic for calculating the cell fill is based on source & target)

Closes #144 

I removed all references to source and target from the app, instead preferring to use _from and _to. This simplifies the logic in data pulling and in aggregation, and keeps it about the same elsewhere. There is one spot that required source and target, the sorting using reorder.js, so there is a spot where source and target are generated as they are needed. The logic to do that already existed, though.

### DIFF
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -301,10 +301,10 @@ export default Vue.extend({
 
       // Count occurrences of links and store it in the matrix
       this.network.links.forEach((link: Link) => {
-        this.matrix[this.idMap[link.source]][this.idMap[link.target]].z += 1;
+        this.matrix[this.idMap[link._from]][this.idMap[link._to]].z += 1;
 
         if (!this.directional) {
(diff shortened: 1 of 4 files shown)


gotdairyya (approved):

mozartfish (dismissed):

waxlamp (approved):
This looks great! In general we should consider whether a computed prop or something similar could be useful for delivering a d3 version of the data to things that need it (this would be more immediately appropriate for multilink, but the same principle could apply here too).

gotdairyya (approved):

JackWilb:
> This looks great! In general we should consider whether a computed prop or something similar could be useful for delivering a d3 version of the data to things that need it (this would be more immediately appropriate for multilink, but the same principle could apply here too).

Opened #165

Issue #150, opened 2020-12-18
Add enter, exit, merge for matrix elements


JackWilb:

JackWilb:
Closed via #167

PR #167, opened 2021-01-05
Add D3 enter, exit, merge to allow the network to be changed in place while triggering a re-render of the matrix

### LINKED ISSUE #148: Sorting an aggregated network causes issues on remove-class branch

JackWilb:
Might address #150 (will need to be tested with a network that will add and remove nodes in place)
Closes #148 

Adds D3 style enter, exit, merge to the rendering function and stops the matrix from being fully removed during each re-render. I also moved some of the icon rendering logic since that doesn't need to be re-rendered on each network update.

Open questions:
- Where should we move the gridlines to? I can see a place for them inside the matrix rows and in the vue rendering. We could also remove them
- Should I try move some of the rendering to the vue layer or do we need D3 for transitions etc.?
- I think I used an old syntax, would it be worth trying to update it to use D3v5 joins instead?

TODO (these can likely just become issues for follow ups):
- [x] Fix lingering bugs with data binding
- [ ] Verify that the computed values for the cells are correct
- [ ] Update attributes when the network updates
- [x] Fix grid line rendering (maybe this is a spot for vue?)
- [x] Sorting is still broken from before (Fixed by resetting the this.matrix variable)
- [ ] Legend seems to get out of sync/doesn't update

### DIFF
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -623,6 +623,7 @@ export default Vue.extend({
     },
 
     drawGridLines(): void {
+      selectAll('.gridLines').remove();
       const gridLines = this.edges.append('g').attr('class', 'gridLines');
 
       const lines = gridLines.selectAll('line').data(this.matrix).enter();
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -226,6 +226,46 @@ export default Vue.extend({
         `translate(${this.visMargins.left},${this.visMargins.top})`,
       );
 
+    // Draw buttons for alternative sorts
+    let initialY = -this.visMargins.left + 10;
+    const buttonHeight = 15;
+    const text = ['name', 'cluster', 'interacts'];
+    const sortNames = ['shortName', 'clusterLeaf', 'edges'];
+    const iconNames = ['alphabetical', 'categorical', 'quant'];
+    for (let i = 0; i < 3; i++) {
+      const button = this.edges
+        .append('g')
+        .attr('transform', `translate(${-this.visMargins.left},${initialY})`);
+      button.attr('cursor', 'pointer');
+      button
+        .append('rect')
+        .attr('width', this.visMargins.left - 5)
+        .attr('height', buttonHeight)
+        .attr('fill', 'none')
+        .attr('stroke', 'gray')
+        .attr('stroke-width', 1);
+      button
+        .append('text')
+        .attr('x', 27)
+        .attr('y', 10)
+        .attr('font-size', 11)
+        .text(text[i]);
+      const path = button.datum(sortNames[i]);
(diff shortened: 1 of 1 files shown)


gotdairyya:
I want to make a case for keeping the gridlines, I think it makes the matrix less overwhelming even though it is more visual "clutter"

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -405,20 +477,46 @@ export default Vue.extend({
...
       this.colorScale
         .domain([0, this.maxNumConnections])
         .range(['#feebe2', '#690000']); // TODO: colors here are arbitrary, change later
 
-      this.edgeRows
+      // Draw cells
+      selectAll('.cellsGroup')
         .selectAll('.cell')
-        .data((d: Node, i: number) => this.matrix[i])
+        .data((d: unknown, i: number) => this.matrix[i])
```
waxlamp, line 512: Cool!

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -226,6 +226,46 @@ export default Vue.extend({
         `translate(${this.visMargins.left},${this.visMargins.top})`,
       );
 
+    // Draw buttons for alternative sorts
+    let initialY = -this.visMargins.left + 10;
+    const buttonHeight = 15;
+    const text = ['name', 'cluster', 'interacts'];
+    const sortNames = ['shortName', 'clusterLeaf', 'edges'];
+    const iconNames = ['alphabetical', 'categorical', 'quant'];
+    for (let i = 0; i < 3; i++) {
```
waxlamp, line 235: A code style suggestion (not required to change it in this PR unless you want to): in cases like this, I think it's better to pack your several arrays' worth of data into a single array of objects, then to call `.forEach()` on that array (or run a for loop against that array's `.length` property). That will help avoid errors you might introduce in the future if you changed the contents of those arrays without updating the `3` in the for loop, etc.

JackWilb: Agreed. I think we should just address this if/when we need to change it

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -240,46 +280,12 @@ export default Vue.extend({
     },
 
     changeMatrix(this: any) {
```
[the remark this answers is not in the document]
JackWilb: Agreed, opened #169

gotdairyya (commented):
resolved my comment, thanks jack!

src/components/MultiMatrix/MultiMatrix.vue
waxlamp: A big reason to make the change is to prevent errors originating from this part of the code, so I'd say it's not a matter of "if" or "when" (no external push will force a fix here). It's more of a weakness in the code that really should be addressed sooner or later (with a preference for sooner).

PR #55, opened 2020-05-14
Add color encoding to cells

gotdairyya:
Closes #49

PR #172, opened 2021-01-05
Fix color scale legend calculation and legend reactivity

### LINKED ISSUE #170: Color scale legend is getting out of sync

JackWilb:
Closes #170

There were a couple issue that needed to be fixed in this PR:

1. The color scale could never reduce its max value, since the max value was never reset
2. Updates to the colorScale didn't automatically trigger an emit, it was a separate event. Using a watcher for this would solve the problem.
3. The code was a little dispersed and could be condensed into a computed prop

I fixed the 3 issues above using the solutions that I mentioned, resetting the max value, using a watcher instead of manually emitting changes to the color scale, and making the colorScale a computed prop.

### DIFF
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -196,6 +196,10 @@ export default Vue.extend({
       this.processData();
       this.changeMatrix();
     },
+
+    colorScale() {
+      this.$emit('updateMatrixLegendScale', this.colorScale);
+    },
   },
 
   async mounted(this: any) {
@@ -274,8 +278,6 @@ export default Vue.extend({
 
     this.initializeAttributes();
     this.initializeEdges();
-
-    this.$emit('updateMatrixLegendScale', this.colorScale);
   },
 
   methods: {
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -62,7 +62,6 @@ export default Vue.extend({
     edges: any;
     edgeColumns: any;
     edgeRows: any;
-    colorScale: ScaleLinear<string, number>;
     icons: { [key: string]: { [d: string]: string } };
     selectedNodesAndNeighbors: { [key: string]: string[] };
     selectedElements: { [key: string]: string[] };
@@ -90,7 +89,6 @@ export default Vue.extend({
       edges: undefined,
       edgeColumns: undefined,
       edgeRows: undefined,
-      colorScale: scaleLinear<string, number>(),
       icons: {
         quant: {
           d:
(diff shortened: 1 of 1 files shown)


gotdairyya (approved):
looks good!

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -176,6 +174,12 @@ export default Vue.extend({
 
       return computedIdMap;
     },
+
+    colorScale(): ScaleLinear<string, number> {
+      return scaleLinear<string, number>()
+        .domain([0, this.maxNumConnections])
+        .range(['#feebe2', '#690000']); // TODO: colors here are arbitrary, change later
```
waxlamp, line 181: File an issue please.

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -291,7 +291,10 @@ export default Vue.extend({
     },
 
     processData(): void {
+      // Reset some values that will be re-calcuated
+      this.maxNumConnections = 0;
```
waxlamp, line 295: Why does this fix the original problem?

waxlamp (approved):
Thanks for pushing these fixes, they look good.

Just a note: in general, whenever you leave a TODO in a code comment that makes it to master, you *must* file an issue to track whatever that TODO is asking for. To make this even easier, you can file the issue first, then link to that issue right in the TODO comment.

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: This was in the code originally, so should have been an issue back when it was merged in #55. I've made an issue to track it, too. See #174

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: If the colorScale updates so that the max value is less, it would fail to update. That's because we would only update the value if it was larger than the previous max. Resetting it to 0 here fixes the issue. Ultimately, we could refactor it into a computed prop that just calculates the largest number of connections instead of computing it manually on each update

src/components/MultiMatrix/MultiMatrix.vue
gotdairyya: That is my mistake @waxlamp, now I know!

src/components/MultiMatrix/MultiMatrix.vue
waxlamp: Ok, gotcha. This seems like a code smell to me because this incidentally means that the other code will happen to work properly. I think the true fix here is to update the "other code" so that it always works properly, rather than only if `maxNumConnections` happens to be set to `0`. I also like your idea of using a computed prop for this.

The specific mental model here borrows from Einstein: "spooky action at a distance", something to generally avoid.

PR #175, opened 2021-01-08
Make the attribute vis reactive to changes

### LINKED ISSUE #168: Attribute label remains persistent
This is a strange bug, to recreate it:

1. Select an attribute
2. Toggle directional edges
3. Remove selected attribute

After these steps, the label remains fixed on top. 

image

JackWilb:
This doesn't address a specific issue, but does seem to fix #168 as a by-product

I copied the same formula from the reactivity of the cells for this and made the attributes reactive. I changed the logic slightly so now it's not done by row, it's instead done as columns, one for each variable, which helps reduce complexity when defining width, x, etc. The zebras and highlights are now much better, too, and they react as you'd expect

Appologies in advance for the large commit 24ed92b, it's a bit of a monster since it was hard to break out the code moving from the logic changes.

### DIFF
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -745,50 +745,31 @@ export default Vue.extend({
       // Just has to be larger than the attributes panel (so that we render to the edge)
       const attributeWidth = 1000;
 
-      this.attributes = select('#attributes')
-        .append('g')
-        .attr('transform', `translate(0,${this.visMargins.top})`);
+      // Add/Update zebras
+      this.attributeZebras = (select('.zebras') as any)
+        .selectAll('.attrRowBackground')
+        .data(this.network.nodes, (d: Node) => d._id || d.id);
 
-      // add zebras and highlight rows
-      this.attributes
-        .selectAll('.highlightRow')
-        .data(this.network.nodes)
-        .enter()
+      this.attributeZebras.exit().remove();
+
+      const attributeZebrasEnter = this.attributeZebras.enter().append('g').attr('class', 'attrRowBackground');
+
+      attributeZebrasEnter
         .append('rect')
-        .classed('highlightRow', true)
-        .attr('x', 0)
+        .classed('zebra', true)
         .attr('y', (d: Node, i: number) => this.orderingScale(i))
         .attr('width', attributeWidth)
         .attr('height', this.orderingScale.bandwidth())
         .attr('fill', (d: Node, i: number) => (i % 2 === 0 ? '#fff' : '#eee'));
 
-      // Draw each row (translating the y coordinate)
-      this.attributeRows = this.attributes
-        .selectAll('.attrRowContainer')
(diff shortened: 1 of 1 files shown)


mozartfish (approved):
Looks good!

gotdairyya (approved):

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -253,6 +253,12 @@ export default Vue.extend({
         `translate(${this.visMargins.left},${this.visMargins.top})`,
       );
 
+    this.attributes = select('#attributes')
+      .append('g')
```
waxlamp, line 261: Couldn't this `g` element just go right in the template? Then you could even set its `transform` attribute with a computed prop, etc. as well.

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -204,11 +204,16 @@ export default Vue.extend({
       });
 
       return scales;
-  },
+    },
+
+    colWidth(): number {
+      const attrWidth = parseFloat(select('#attributes').attr('width'));
```
waxlamp, line 208: It *seems* like the `width` attribute of `#attributes` should be a data prop in some capacity, which would in turn set that attribute on `#attributes` in the template (see my previous comment about this as well), and also drive the computation here.

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -831,9 +842,36 @@ export default Vue.extend({
...
+        .attr('d', (d: string) => {
+          const type = this.isQuantitative(d) ? 'quant' : 'categorical';
+          return this.icons[type].d;
+        })
+        .attr(
+          'transform',
+          (d: string, i: number) =>
+            `scale(0.1)translate(${
+              (this.colWidth + this.colMargin) * i * 10 - 200
+            }, -1100)`,
```
waxlamp, line 855: This is somewhat hard to read. It would benefit from restructuring into a non-expression function, but that's totally an optional change at this point.

waxlamp (approved):
Approving, though I'd look at my comments about driving some of the attributes with Vue props and see if they're valid. If so, you can just file appropriate issues and defer fixing until an appropriate time.

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: Opened #179

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: Quoted in #179, since it seems like it could be taken care of in one change set

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: This is a prettier solution to a problem that didn't really exist. I'll make a note of this in that PR and then I'll update it there before I merge it in

PR #176, opened 2021-01-08
Move the matrix sizing to the template

JackWilb:
This doesn't address a specific issue

The matrix was not resizing and since we're adding and removing rows, that's become an issue. This moves the resizing code from the `mounted()` function to the template so that the size is reactive to changing numbers of nodes. I also migrated the sizing for the attributes panel so now they should all stay in sync


mozartfish (approved):
Looks good!

gotdairyya (approved):
LGTM!

waxlamp (commented):
Good improvement.

PR #177, opened 2021-01-09
Make the cells reactive to updates from prop

### LINKED ISSUE #171: Cell fills are not reactive to directional edges, tooltip is
This looks like a bug introduced by #167

JackWilb:
Closes #171

The code that renders the cells was not able to update existing DOM elements with new fill and with new position (e.g. @mozartfish updates). This fix solves that problem by defining the fill and position on the new elements. To do that, I need to add a component level data variable to track the cells (this.cells) and then call the updates on that.


mozartfish (approved):
Looks good to me!

gotdairyya (approved):

waxlamp (approved):

PR #188, opened 2021-01-25
Animating Rows - Fixes Issue #181

### LINKED ISSUE #181: Fix animation of rows that are exiting and entering from aggregated matrix
The current start point for the enter animation is wrong, it should be the super node, not 0,0, and the rows don't animate on exit.

mozartfish:
Close #181 
Adds index tracking code to allow for children nodes to drop down from their parent node positions. Exit Transition was not fixed in this PR because of some timing issue with Vue. For column labels, a fade transition was added as well to animate the columns.

### DIFF
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -430,12 +430,12 @@ export default Vue.extend({
           return `translate(${this.orderingScale(i)})rotate(-90)`;
         });
 
-      // this.edgeColumns.exit().remove();
-      this.edgeColumns.exit()
-      .transition()
-      .duration(1000)
-      .style('opacity', 0.2)
-      .remove();
+      this.edgeColumns
+        .exit()
+        .transition()
+        .duration(1000)
+        .style('opacity', 0.2)
+        .remove();
 
       const columnEnter = this.edgeColumns
         .enter()
@@ -518,11 +518,7 @@ export default Vue.extend({
           return `translate(0,${this.orderingScale(i)})`;
         });
 
-      // this.edgeRows.exit().remove();
       this.edgeRows.exit()
-      .transition()
-      .duration(1000)
-      .style('opacity', 0.2)
       .remove()
 
       const rowEnter = this.edgeRows
@@ -613,7 +609,11 @@ export default Vue.extend({
       // Draw cells
       this.cells = selectAll('.cellsGroup')
         .selectAll('.cell')
-        .data((d: unknown, i: number) => this.matrix[i]);
+        .data((d: unknown, i: number) => {
+          console.log(i, this.matrix[i]);
+          return this.matrix[i];
+        });
+      // .data((d: unknown, i: number) => this.matrix[i]);
 
       // Update existing cells
       this.cells
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -442,7 +442,7 @@ export default Vue.extend({
               d.parentPosition,
             )})rotate(-90)`;
           } else {
-            return `translate(0, 0)`;
+            return `translate(0, 0)rotate(-90)`;
           }
         });
 
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -430,12 +430,7 @@ export default Vue.extend({
           return `translate(${this.orderingScale(i)})rotate(-90)`;
         });
 
-      this.edgeColumns
-        .exit()
-        .transition()
-        .duration(500)
-        .style('opacity', 0.2)
-        .remove();
+      this.edgeColumns.exit().remove()
 
       const columnEnter = this.edgeColumns
         .enter()
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -518,7 +518,12 @@ export default Vue.extend({
           return `translate(0,${this.orderingScale(i)})`;
         });
 
-      this.edgeRows.exit().remove();
+      // this.edgeRows.exit().remove();
+      this.edgeRows.exit()
+      .transition()
+      .duration(1000)
+      .style('opacity', 0.2)
+      .remove()
 
       const rowEnter = this.edgeRows
         .enter()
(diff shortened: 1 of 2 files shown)


gotdairyya:
1. Seems like there is a bug where when you try to "open" more than one row it doesn't work (this is a bug that sometimes occurs and sometimes doesn't, @JackWilb does this happen for you too?)
2. I'm not the biggest fan of the enter transitions (but this is not an important issue) I think it would be cleaner if the labels just occurred in place

JackWilb:
>     1. Seems like there is a bug where when you try to "open" more than one row it doesn't work (this is a bug that sometimes occurs and sometimes doesn't, @JackWilb does this happen for you too?)

I'm not seeing that behavior. Do you have a reproducible example? Can you share a video of it?

>     2. I'm not the biggest fan of the enter transitions (but this is not an important issue) I think it would be cleaner if the labels just occurred in place

Once we have the animation, it would be easy to decide what would be the best visually. We'll likely want input from Alex and Jared on this, so let's not worry to much about this for now. We could make an issue out of it, if you think it's a big deal

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -430,7 +430,12 @@ export default Vue.extend({
           return `translate(${this.orderingScale(i)})rotate(-90)`;
         });
 
-      this.edgeColumns.exit().remove();
+      this.edgeColumns
+        .exit()
+        .transition()
+        .duration(500)
+        .style('opacity', 0.2)
+        .remove();
```
JackWilb, line 438: Neither the rows nor columns seem to actually animate out, they just pop. Is this code actually doing anything? Do you need to add similar code to more places for a better animation?

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -519,11 +524,17 @@ export default Vue.extend({
...
         .append('g')
         .attr('class', 'rowContainer')
-        .attr('transform', `translate(0, 0)`);
+        .attr('transform', (d: Node) => {
+          if (d.type === 'node') {
+            return `translate(0, ${this.orderingScale(d.parentPosition)})`;
+          } else {
+            return `translate(0, 0)`;
+          }
+        });
```
JackWilb, line 533: If we're going to do this for the rows, does it make sense to also do it for the columns? I'm just wondering how that would look overall

src/lib/aggregation.ts
```diff
@@ -211,15 +214,32 @@ function expandSuperNodeData(
...
-      superNodeCopy.splice(superIndexStart + count, 0, node);
+      nodeCopy.splice(insertNodeStart + count, 0, node);
       count += 1;
     });
 
-    return superNodeCopy;
+    // Update the index of the new nodes the expanded vis network
+    nodeCopy.forEach((node: Node, index: number) => {
+      node.index = index;
+    });
```
JackWilb, line 228: Is this used anywhere? I don't think find index needs and index property on the objects

src/lib/aggregation.ts
```diff
@@ -363,6 +384,10 @@ function retractSuperNodeData(
     const superIndexFunc = (superNode: Node) => superNode.id == superNodeName;
     const superIndexStart = expandNodesCopy.findIndex(superIndexFunc);
     expandNodesCopy.splice(superIndexStart + 1, childNodes.length);
+    // Update the index of the new nodes in the retracted vis network
+    expandNodesCopy.forEach((node: Node, index: number) => {
+      node.index = index;
+    });
```
JackWilb, line 390: Same as above

JackWilb (changes requested):

JackWilb:
This really breaks sorting, by the way. Please file an issue that we need to fix that

gotdairyya:
> > ```
> > 1. Seems like there is a bug where when you try to "open" more than one row it doesn't work (this is a bug that sometimes occurs and sometimes doesn't, @JackWilb does this happen for you too?)
> > ```
> 
> I'm not seeing that behavior. Do you have a reproducible example? Can you share a video of it?
> 
Tried to replicate and it is working fine now, not sure what happened there.
> > ```
> > 2. I'm not the biggest fan of the enter transitions (but this is not an important issue) I think it would be cleaner if the labels just occurred in place
> > ```
> 
> Once we have the animation, it would be easy to decide what would be the best visually. We'll likely want input from Alex and Jared on this, so let's not worry to much about this for now. We could make an issue out of it, if you think it's a big deal

Nope, no need to file an issue.

src/lib/aggregation.ts
mozartfish: resolved with this commit https://github.com/multinet-app/multimatrix/pull/188/commits/e1bb54b43a69d9a9953df7a107da73ebf0698057

src/lib/aggregation.ts
mozartfish: resolved with this commit https://github.com/multinet-app/multimatrix/pull/188/commits/304c01d3c9465e7aaf196633c42b9b40eb95063c

src/components/MultiMatrix/MultiMatrix.vue
mozartfish: If you look very carefully there is a slightly fast fade for the column labels. I will take these out since I haven't figured out the row exit and I will file an issue since this transition seems like it can be fixed later once I have everything else working

mozartfish: removed column transition with this commit https://github.com/multinet-app/multimatrix/pull/188/commits/00641546626d6f7684249ef57df52a9750c5c155

src/components/MultiMatrix/MultiMatrix.vue
mozartfish: The transition is super cool and applied it with this commit https://github.com/multinet-app/multimatrix/pull/188/commits/13b1fe916b2ebc81e533fa2ece63fe6e749d6307

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -436,6 +436,19 @@ export default Vue.extend({
         .enter()
         .append('g')
         .attr('class', 'column')
+        .attr('transform', (d: Node) => {
+          if (d.type === 'node') {
+            return `translate(${this.orderingScale(
+              d.parentPosition,
+            )})rotate(-90)`;
+          } else {
+            return `translate(0, 0)`;
```
JackWilb, line 445: This needs to be rotated
```suggestion
            return `translate(0, 0)rotate(-90)`;
```

JackWilb (changes requested):
Just one minor thing I noticed

JackWilb (approved):

PR #192, opened 2021-02-01
Indent Child Node Row Labels and Create Visual Identifier for Child Node Row and Column Labels

### LINKED ISSUE #184: Show row hierarchy between parent and children nodes (labels)

### LINKED ISSUE #185: Show column hierarchy between parent and children nodes (labels)

### LINKED ISSUE #193: Fix Child Node Row Label Positioning
The child row labels indentation overlap with the matrix for the Les Mis Data and datasets larger than the flight data
image

mozartfish:
Close #184 
Close #185 
Close #193

### DIFF
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -523,9 +523,6 @@ export default Vue.extend({
         .selectAll('.rowContainer')
         .data(this.network.nodes, (d: Node) => d._id || d.id)
         .attr('transform', (d: Node, i: number) => {
-          if (d.type === 'node') {
-            return `translate(10, ${this.orderingScale(i)})`;
-          }
           return `translate(0,${this.orderingScale(i)})`;
         });
 
@@ -537,7 +534,7 @@ export default Vue.extend({
         .attr('class', 'rowContainer')
         .attr('transform', (d: Node) => {
           if (d.type === 'node') {
-            return `translate(10, ${this.orderingScale(d.parentPosition)})`;
+            return `translate(0, ${this.orderingScale(d.parentPosition)})`;
           } else {
             return `translate(0, 0)`;
           }
@@ -547,9 +544,6 @@ export default Vue.extend({
         .transition()
         .duration(1000)
         .attr('transform', (d: Node, i: number) => {
-          if (d.type === 'node') {
-            return `translate(10, ${this.orderingScale(i)})`;
-          }
           return `translate(0,${this.orderingScale(i)})`;
         });
 
@@ -569,7 +563,15 @@ export default Vue.extend({
       // add foreign objects for label
       rowEnter
         .append('foreignObject')
-        .attr('x', -rowLabelContainerStart)
+        .attr('x', (d: Node) => {
+          if (d.type === 'node') {
+            return (-rowLabelContainerStart + 15)
+          }
+          else {
+            return (-rowLabelContainerStart);
+          }
+        })
+        // .attr('x', -rowLabelContainerStart)
         .attr('y', -5)
         .attr('width', labelContainerWidth)
         .attr('height', labelContainerHeight)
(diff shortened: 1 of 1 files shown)


src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -489,6 +489,14 @@ export default Vue.extend({
...
         });
 
+      columnEnter.selectAll('p').style('color', (d: Node) => {
+        if (d.type === 'node') {
+          return '#aaa';
+        } else {
+          return 'black';
+        }
+      });
+
```
JackWilb, line 499: Instead of using a new block here, can we add this style setting near line 465? I would imagine you can just do the `.style()` instead of having to do the whole selection again

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -563,13 +571,29 @@ export default Vue.extend({
...
-        .attr('x', -rowLabelContainerStart)
+        .attr('x', (d: Node) => {
+          if (d.type === 'node') {
+            return -rowLabelContainerStart + 15;
+          } else {
+            return -rowLabelContainerStart;
+          }
+        })
         .attr('y', -5)
         .attr('width', labelContainerWidth)
```
JackWilb, line 582: You can solve #193 by modifying this width here

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -563,13 +571,29 @@ export default Vue.extend({
...
-        .classed('rowLabels', true)
+        .classed('rowLabels', true);
+
+      rowEnter.selectAll('p').style('color', (d: Node) => {
+        if (d.type === 'node') {
+          return '#aaa';
+        } else {
+          return 'black';
+        }
+      });
```
JackWilb, line 594: Same as above. Do we need another code block for this?

JackWilb (changes requested):

src/components/MultiMatrix/MultiMatrix.vue
mozartfish: Close #193 https://github.com/multinet-app/multimatrix/pull/192/commits/1485821a03126e30d871d36d2c087152fac0265d

src/components/MultiMatrix/MultiMatrix.vue
mozartfish: we need this to be separate because first selection deals specifically with the foreign object which has the paragraph element added to it

JackWilb: This style is being applied to the p tag in the foreign object though, right? If so, this code should live with that

mozartfish: Do you mean add the styling to the CSS definition?

JackWilb: I mean above, where you have

```
        .append('xhtml:p')
        .text((d: Node) => d._key)
        .classed('colLabels', true)
```

replace it with

```
        .append('xhtml:p')
        .text((d: Node) => d._key)
        .style('color', (d: Node) => {
          if (d.type === 'node') {
            return '#aaa';
          } else {
            return 'black';
          }
        })
        .classed('colLabels', true)
```

JackWilb: Instead of making a new selection

mozartfish: this change produces this 
image

mozartfish: the child labels are black in this picture instead of grey

JackWilb: That image is from the wrong branch. We don't have node coloring here

JackWilb: I pushed the change to fix it. If it's working on your end, will you resolve these 2 open threads

src/components/MultiMatrix/MultiMatrix.vue
JackWilb: Moved your close statement to the PR description

JackWilb (approved):

PR #199, opened 2021-02-11
Table attributes

### LINKED ISSUE #197: Fix Table Attribute Grouping
Table is broken for aggregation and needs to be fixed to display the distinct groups collapsed 
image

mozartfish:
Close #197 
This commit fixes a bug for the attribute table. The bug grouped all the aggregate nodes into a single category when the matrix was grouped in a supernode such as in this image from the multimatrix.app page 
image

The fix is demonstrated below
https://user-images.githubusercontent.com/24800816/107675699-3147c200-6c4d-11eb-8ae1-a4192f819af0.mp4

Remaining issue is #200

### DIFF
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -235,9 +235,14 @@ export default Vue.extend({
           scale.clamp(true);
           scales[col] = scale;
         } else {
-          const values: string[] = this.network.nodes.map(
-            (node: Node) => node[col],
-          );
+          const values: string[] = [];
+          this.network.nodes.forEach((node: Node) => {
+            if (node.type === 'supernode') {
+              values.push(node['GROUP']);
+            } else {
+              values.push(node[col]);
+            }
+          });
           const domain = [...new Set(values)];
           const scale = scaleOrdinal(schemeCategory10).domain(domain);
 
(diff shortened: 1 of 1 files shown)


gotdairyya (approved):

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -235,9 +235,14 @@ export default Vue.extend({
...
-            (node: Node) => node[col],
-          );
+          const values: string[] = [];
+          this.network.nodes.forEach((node: Node) => {
+            if (node.type === 'supernode') {
+              values.push(node['GROUP']);
+            } else {
+              values.push(node[col]);
+            }
+          });
```
JackWilb, line 245: This should still be a map, since for each value in .nodes you're getting a new value.

JackWilb (changes requested):
This is a little code format thing I noticed. I'll be able to give you a more thorough review once I have a bit more info

src/components/MultiMatrix/MultiMatrix.vue
mozartfish: I tried using map but ran into issues with how to apply the type checking since supernodes group attribute is the same as the node[col] attribute

JackWilb: What do you mean by apply type checking?

Just do this:
```
          const values: string[] = this.network.nodes.map((node: Node) => {
            if (node.type === 'supernode') {
              return node['GROUP'];
            } else {
              return node[col];
            }
          });
```

mozartfish: this makes more sense. By type checking I meant typechecking for the nodes (supernode, undefined, childnode)

mozartfish: fixed with this commit https://github.com/multinet-app/multimatrix/pull/199/commits/5b4f6eb0f6caac37ee41e27df4f37935d0271d45

JackWilb (approved):

PR #206, opened 2021-02-16
Allow for a larger node table (upto 300 nodes per table)

### LINKED ISSUE #204: Increase MultiMatrix table load size
MultiMatrix table size needs to be increased to enable larger tables like Les Mis and Flights can be displayed

JackWilb:
Closes #204

Bumps the download limit for a node table to 300 nodes. This make the app work with the boston dataset, again.


gotdairyya (approved):
Looks good!

PR #217, opened 2021-03-15
Use the vuex store for data fetching

JackWilb:
Converted to draft. Will merge after Pranav's work

This copies over some of the logic from the MultiLink application for pulling data using the `multinetjs` library. Since we're now using the API types, I modified some of the types in the application so that we're aligned with the API. This removes some duplicated logic that would check for `id || _id` and some functions that were duplicated for each case.

One knock-on effect of this change is that we no longer need to do an `emit()` when the network changes, since we can just set the network to the updated value using the store. An additional change we can make here, is to use the store for tracking the other objects that are emitted to simplify that logic (I'll open an issue)

TODO:
- [x] Open an issue to track moving emitted object to store (#220)
- [x] Open an issue to stop the data binding, preferring using the store instead (#221)
- [x] Data fetching should be triggered on App mount (#233)


JackWilb:
Currently a bug with neighbor highlights since neighbor generation has been removed

JackWilb:
This is change implements similar functionality to the MultiLink app, so I'm going to forego a review round here. If we have any gripes with the implementation, we'll now need to change it in both places

PR #222, opened 2021-03-22
Restore Matrix

### LINKED ISSUE #164: Add toggle for aggregation
Create a toggle that allows users to switch back and forth between viewing an aggregated version of the matrix and the original matrix

mozartfish:
This PR implements restoring the matrix to its original non-aggregated state when the user toggles the enable graffinity features to the left. This PR builds upon the following PR and its associated issues: Aggregation UI Buttons

Close #164 

Demo 

https://user-images.githubusercontent.com/24800816/112214935-02722380-8bdd-11eb-9fb6-a2b0ddcfcad2.mp4

### DIFF
--- a/src/components/MultiMatrix/MultiMatrix.vue
+++ b/src/components/MultiMatrix/MultiMatrix.vue
@@ -95,6 +95,7 @@ export default Vue.extend({
     sortKey: string;
     colMargin: number;
     showIcon: boolean;
+    disableAggr: boolean;
   } {
     return {
       browser: {
@@ -154,6 +155,7 @@ export default Vue.extend({
       sortKey: '',
       colMargin: 5,
       showIcon: false,
+      disableAggr: false,
     };
   },
 
@@ -295,7 +297,7 @@ export default Vue.extend({
     },
 
     enableGraffinity() {
-      if (this.enableGraffinity === false) {
+      if (this.enableGraffinity === false && this.disableAggr === true) {
         console.log('Disabled Aggregation!');
 
         // Clear the click map so correct icons are drawn for aggregation
@@ -1247,6 +1249,9 @@ export default Vue.extend({
               'updateNetwork',
               superGraph(this.network.nodes, this.network.links, d),
             );
+
+            // Turn on the disable aggregation
+            this.disableAggr = true;
 
             // View/Hide Matrix Legends
             this.$emit('updateMatrixLegends', true, false);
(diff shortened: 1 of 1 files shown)


JackWilb (changes requested):
Looks good, code-wise. There is a bug though. If you enable graffinity features and then disable them without aggregating, it deletes the network

mozartfish:
Fix the disable aggregation error https://github.com/multinet-app/multimatrix/pull/222/commits/586e5aefc6c6440c68e6d1307b85fefcbae81c72

JackWilb (changes requested):
Let's call this variable `aggregated`. You need to initialize it as false, set it to true when the network aggregation is called, and then set it back to false once the matrix has been reset to its initial state. 

Another quick note, you don't need to do `this.enableGraffinity === false`, just do `!this.graffinity`

mozartfish:
Fix variable names to aggregated and clean up code with this commit https://github.com/multinet-app/multimatrix/pull/222/commits/5c0bceff082aeb153db3a5d0878e7755a941e434

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -292,6 +295,60 @@ export default Vue.extend({
       this.processData();
       this.changeMatrix();
     },
+
+    enableGraffinity() {
+      if (!this.enableGraffinity && this.aggregated === true) {
+        console.log('Disabled Aggregation!');
+
+        // Clear the click map so correct icons are drawn for aggregation
+        this.clickMap.clear();
```
JackWilb, line 304:
```suggestion
      if (!this.enableGraffinity && this.aggregated === true) {
        // Clear the click map so correct icons are drawn for aggregation
        this.clickMap.clear();
```

src/components/MultiMatrix/MultiMatrix.vue
```diff
@@ -292,6 +295,60 @@ export default Vue.extend({
...
+          .data(this.network.nodes, (d: Node) => d._id || d.id)
+          .attr('y', -5)
+          .attr('x', columnLabelContainerStart)
+          .attr('width', labelContainerWidth)
+          .attr('height', labelContainerHeight);
+
+        (selectAll('.colLabels') as any)
+          .data(this.network.nodes, (d: Node) => d._id || d.id)
+          .style('color', 'black')
+          .classed('rowLabels', true);
```
JackWilb, line 343: This is in the wrong spot. It should be a part of the render method. I can move this if you'd prefer that

JackWilb (changes requested):
Looks good overall. The code is in the wrong place, though. It should be in the rendering method. I'm happy to fix that if you can't figure it out

src/components/MultiMatrix/MultiMatrix.vue
mozartfish: If you can move this section that would be really helpful. Thank you!

JackWilb (approved):
Actually, I'm fine with where this is. Combining it into the render function will be significant work and this is good for now

PR #242, opened 2021-04-29
Add lineupJS as our attribute visualization

JackWilb:
Add lineupJS as our node attribute visualization

To resolve the compilation errors with lineup, I had to `skipLibCheck` in the tsconfig. This means that we no longer type check the imported libs, reducing our type safety slightly. See #258 

TODO:
- [x] Map highlights between the two 
- [x] Type errors in lineup cause the app not to build
- [x] Add link attributes to vis (@gotdairyya, can you help me take a look at this?) (#264)
- [x] Fix row flex width
- [x] Fix height of the lineup
- [x] Fix header alignment
- [x] Track when header height changes and make the matrix track it (on grouping, etc.) (#261)
- [x] Map sorting between the two (#260)
- [x] Fix sorting #259 
- [x] Fix filtering #262


gotdairyya (dismissed):
I created a few issues with some of the bugs that I could catch. Other than that, it is running locally and looks good!

PR #246, opened 2021-05-06
Subset large networks

### LINKED ISSUE #207: Add support for visualizing large table subset
In instances where there is too much data to display, prompt the user to subset the data either by choosing:

1. a node attribute to aggregate by
2. the top X nodes based on connectivity
3. X random sampling of nodes

gotdairyya:
Closes #207 

This branch creates a dialog for users to subset large networks. There are two options for subsetting:
1. Choosing a max number of edges to visualize
2. Choosing a node attribute to aggregate by

**To Do:**

- [ ] Add subset option to aggregate by categorical attribute
- [ ] Make it look nicer
- [ ] Discuss and determine whether the current AQL query is appropriate for subsetting
- [x] Add a catch for IF the AQL query fails


src/components/Alert.vue
```diff
@@ -114,6 +120,13 @@ export default {
         </v-col>
       </v-row>
     </v-alert>
+
+    <filter-overlay
+      v-if="
+        loadError.message ===
+          'The network you are loading is too large'
+      "
+    />
```
JackWilb, line 129:
```suggestion
    <filter-overlay
      v-if="loadError.message === 'The network you are loading is too large'"
    />
```

src/store/index.ts
```diff
@@ -64,6 +65,9 @@ const {
         .domain([0, state.maxConnections.child])
         .range(['#f79d97', '#c0362c']);
     },
+    nodeTableName(state: State) {
+      return state.nodeTableName;
+    },
```
JackWilb, line 70: I've just learned that we don't need to make these getters that return the value like this (I have them removed in the lineup branch). Instead we should delete it and use `store.state.nodeTableName` when we're using the value elsewhere

```suggestion
```

src/store/index.ts
```diff
@@ -177,7 +184,22 @@ const {
       if (networkTables === undefined) {
         return;
       }
+      // Check node and table size
```
JackWilb, line 187: It's a bit cleaner if we add a line of white space above this block
```suggestion
     
      // Check node and table size
```

src/store/index.ts
```diff
@@ -177,7 +184,22 @@ const {
...
+      const sizePromises = networkTables.nodeTables.map((table) => api.aql(workspaceName, `FOR doc IN ${table} COLLECT WITH COUNT INTO length RETURN length`));
 
+      const resolvedSizePromises = await Promise.all(sizePromises);
```
JackWilb, line 190:
```suggestion
      const sizePromises = networkTables.nodeTables.map((table) => api.aql(workspaceName, `FOR doc IN ${table} COLLECT WITH COUNT INTO length RETURN length`));
      const resolvedSizePromises = await Promise.all(sizePromises);
```

src/store/index.ts
```diff
@@ -177,7 +184,22 @@ const {
...
+          commit.setLoadError({
+            message: 'The network you are loading is too large',
+            href: 'https://multinet.app',
+          });
+        }
+      });
+      commit.setNodeTableName(networkTables.nodeTables[0]);
+      if (store.state.loadError.message !== '') {
+        return;
+      }
```
JackWilb, line 202:
```suggestion
      }

```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+      <v-row
+        align="center"
+        justify="center"
+      >
+        <v-overlay
+          absolute
+          :value="overlay"
+        >
+          <v-card>
+            <v-list>
```
JackWilb, line 13: Is there a specific reason you chose a list here, or it was just in the examples you found?

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
+<template>
+  <div id="dialog">
+    <v-app id="filterOverlayDialog">
```
JackWilb, line 3: We only really need IDs to reference them elsewhere
```suggestion
  <div>
    <v-app>
```

JackWilb, line 3: I'd recommend not using v-app here. It could cause some weird css issues. Let's just remove it, since the div is a good containing element

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+import store from '@/store';
+
+Vue.use(Vuex);
+
+export default Vue.extend({
+  data: () => ({
+    overlay: true,
+    zIndex: 0,
+    min: 10,
+    max: 300,
```
JackWilb, line 79: You don't need to make these data props if you're just using them once in the template. Just put the number right in the template, e.g. `:max="300"`. Same for overlay, max, and min.

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+            <v-card>
+              <v-divider />
+              <v-card-text>
+                <div>Choose a random subset of edges</div>
+                <v-slider
+                  v-model="subsetAmount"
+                  :max="max"
+                  :min="min"
+                  step="10"
+                  :ticks="true"
```
JackWilb, line 32:
```suggestion
                  ticks
```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+  Network, Link,
+} from '@/types';
+import store from '@/store';
+
+Vue.use(Vuex);
+
+export default Vue.extend({
+  data: () => ({
+    overlay: true,
+    zIndex: 0,
```
JackWilb, line 77: This appears to be unused
```suggestion
```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+              <v-card-text>
+                <div>Choose a random subset of edges</div>
+                <v-slider
+                  v-model="subsetAmount"
+                  :max="max"
+                  :min="min"
+                  step="10"
+                  :ticks="true"
+                  thumb-label
+                >
```
JackWilb, line 34: This will make the label visible
```suggestion
                  thumb-label
                  thumb-color="primary"
                >
```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+                    style="width: 60px"
+                  />
+                </v-slider>
+              </v-card-text>
+              <v-divider />
+            </v-card>
+
+            <v-btn
+              class="white--text"
+              color="teal"
```
JackWilb, line 50: Take this or leave this. I think Jared has a better idea of what to do, but this will make the button blue
```suggestion
              color="primary"
```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+              @click="filterNetwork"
+            >
```
JackWilb, line 52: `tile` removes the radius from the button. It's still slightly too large though on my machine with this change
```suggestion
              @click="filterNetwork"
              tile
            >
```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+import Vue from 'vue';
+import Vuex from 'vuex';
+import api from '@/api';
+import {
+  Network, Link,
+} from '@/types';
+import store from '@/store';
+
+Vue.use(Vuex);
+
```
JackWilb, line 73:
```suggestion

```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+        </v-overlay>
+      </v-row>
+    </v-app>
+  </div>
+</template>
+
+<script lang="ts">
+/* eslint-disable no-restricted-syntax */
+import Vue from 'vue';
+import Vuex from 'vuex';
```
JackWilb, line 65:
```suggestion
```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+    workspace() { return store.state.workspaceName; },
+  },
+
+  methods: {
+    filterNetwork() {
+      if (this.workspace === null) {
+        return;
+      }
+
+      this.overlay = false;
```
JackWilb, line 94: Maybe do this after the function succeeds? What if it fails here?

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+            </v-btn>
+          </v-card>
+        </v-overlay>
+      </v-row>
+    </v-app>
+  </div>
+</template>
+
+<script lang="ts">
+/* eslint-disable no-restricted-syntax */
```
JackWilb, line 63: I have suggestions for how to fix this below
```suggestion
```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+              arr.id = arr._id;
+              aqlNetwork.nodes.push(arr);
+              nodeChecker.splice(nodeChecker.indexOf(arr._key), 1);
+            }
+          }
+          path.edges.map((arr: Link) => aqlNetwork.edges.push(arr));
+        }
+        // Update state with new network
+        store.commit.setNetwork(aqlNetwork);
+      });
```
JackWilb, line 123: You'd asked for my help refactoring this. Here's what I came up with. I used a set to track the `_id`, instead of key since that relies on the table. If you had nodes from 2 different tables with the same key, you might have had problems, and a set makes the logic much cleaner.
```suggestion
      newTablePromise.then((promise) => {
        const aqlNetwork: Network = { nodes: [], edges: [] };

        // check for duplicates
        const nodeChecker: Set<string> = new Set();
        promise.forEach((path) => {
          path.vertices.forEach((node: Node) => {
            nodeChecker.add(node._id);
          });
        });

        // create network in json format
        promise.forEach((path) => {
          path.vertices.forEach((node: Node) => {
            if (nodeChecker.has(node._id)) {
              aqlNetwork.nodes.push(node);
              nodeChecker.delete(node._id);
            }
          });
          aqlNetwork.edges = path.edges;
        });

        // Update state with new network
        store.commit.setNetwork(aqlNetwork);
      });
```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,128 @@
...
+  },
+
+  methods: {
+    filterNetwork() {
+      if (this.workspace === null) {
+        return;
+      }
+
+      this.overlay = false;
+      const aqlQuery = `FOR nodes in ${this.nodeTable} LIMIT ${this.subsetAmount} FOR v,e,p in 1..4 ANY nodes GRAPH '${this.workspace}' LIMIT ${this.subsetAmount} RETURN p`;
```
JackWilb, line 95: I think `GRAPH '${this.workspace}'` is a bug

JackWilb (changes requested):
Looks great so far, even though I've left a ton of comments. Some of them are just for general style and simplification and I've left a larger change where you asked for my help with refactoring. Ask questions where things aren't clear 😄

src/components/FilterOverlay.vue
JackWilb: If you accept this change, you'll also need to mess with the imports to add `Node`

src/components/Alert.vue
```diff
@@ -1,12 +1,16 @@
...
 } from '@vue/composition-api';
 import api from '@/api';
+import FilterOverlay from '@/components/FilterOverlay.vue';
 
-export default {
-  name: 'Alert',
+export default Vue.extend({
+  components: {
+    FilterOverlay,
+  },
```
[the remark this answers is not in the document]
gotdairyya: Ok good to know, thank you!

src/components/Alert.vue
```diff
@@ -30,11 +34,15 @@ export default {
 
     const buttonHref: Ref<string> = ref(loadError.value.href);
     const buttonText: Ref<string> = ref('');
+    // Add button
```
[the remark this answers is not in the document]
gotdairyya: Oh yes that makes total sense

src/components/FilterOverlay.vue
gotdairyya: Yes, there is going to be one more option for how to subset (hence the list, but maybe it's not necessary?)

src/components/FilterOverlay.vue
gotdairyya: No, I think this is the right thing to do (plus we can ask Jared)

src/components/FilterOverlay.vue
gotdairyya: What do you mean by "this" in the first part of your comment?
And yes, I will add what happens if it fails

src/components/FilterOverlay.vue
gotdairyya: what do you mean?

src/components/FilterOverlay.vue
gotdairyya: A set is so clever! and yes this refactoring is much easier to follow -- thank you!

src/store/index.ts
gotdairyya: oh, so we're completely removing all the getters and then we can still use `store.state.nodeTableName` ?

src/components/FilterOverlay.vue
JackWilb: I was thinking the overlay shouldn't hide until it's done processing

src/store/index.ts
JackWilb: Yeah, the getters should only be used if you need to do something like adding 2 different values from the store (and only if it's used across multiple components)

src/components/FilterOverlay.vue
gotdairyya: Don't I need to keep `overlay` as a data prop though because it needs to be updated on click

src/components/FilterOverlay.vue
gotdairyya: Oh oh, you are so right.

src/components/FilterOverlay.vue
gotdairyya: Resolved with c7ca8ee84382529a376ec2ea73171a6c4c7c39f3

src/components/FilterOverlay.vue
gotdairyya: Resolved but slightly modified. Since there can be multiple edges in a path I add a map. Closed with c7ca8ee84382529a376ec2ea73171a6c4c7c39f3

gotdairyya:
> You should consider opening an issue to convert this code to composition API style (I'm trying to write all our new components in that style), and you should think about catching errors with the AQL and how that will work.

New issue opened #247 for composition API. I may need help with it since some of it still doesn't make sense.

src/components/Alert.vue
```diff
@@ -1,13 +1,19 @@
 <script lang="ts">
+import Vue from 'vue';
```
JackWilb, line 2: Is this unused now?

src/components/FilterOverlay.vue
JackWilb: I don't think this matters, for now. So long as it works :)

src/components/FilterOverlay.vue
JackWilb: Looks like this wasn't fixed

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,122 @@
...
+
+        // create network in json format
+        promise.forEach((path) => {
+          path.vertices.forEach((node: Node) => {
+            if (nodeChecker.has(node._id)) {
+              aqlNetwork.nodes.push(node);
+              nodeChecker.delete(node._id);
+            }
+          });
+          path.edges.map((edge: Link) => aqlNetwork.edges.push(edge));
```
JackWilb, line 112: This shouldn't be any different from 
```suggestion
          aqlNetwork.edges = path.edges;
```
All you're doing is pushing the edge for each edge. Instead of looping and pushing, we can just assign

JackWilb (changes requested):
Another couple of suggestions, but this is looking almost ready to go

src/components/FilterOverlay.vue
JackWilb: Instead of having the `overlay`, you could just update the `loadError` and it would unmount the component

src/components/FilterOverlay.vue
gotdairyya: No, it is different, because every time you loop through the path it replaces the edges in aqlNetwork, so you only end up with the last set of paths

src/components/Alert.vue
gotdairyya: yes, closed with 3d4febd0ec788f85fca2a70b17c7ef0c4ed89fbf

src/components/FilterOverlay.vue
JackWilb: Updated with spread

src/components/FilterOverlay.vue
gotdairyya: fixed with updating the state: a1075fc2cbe0cad2c604b134dc95d0a19eb74846

src/components/FilterOverlay.vue
gotdairyya: Removed with 306ec78fe825ed4c453b89b8a05b982b7a755bf7

src/components/FilterOverlay.vue
gotdairyya: Closed with a1075fc2cbe0cad2c604b134dc95d0a19eb74846

src/components/FilterOverlay.vue
gotdairyya: Resolved with a1075fc2cbe0cad2c604b134dc95d0a19eb74846

src/components/FilterOverlay.vue
gotdairyya: Resolved with a1075fc2cbe0cad2c604b134dc95d0a19eb74846
Turns out the i should've been call the network name after `GRAPH`

JackWilb (changes requested):
We're getting there. Just a couple more issues to address with the AQL query and catching the errors it could have

src/store/index.ts
```diff
@@ -88,6 +89,9 @@ const {
     setUserInfo(state, userInfo: UserSpec | null) {
       state.userInfo = userInfo;
     },
+    setNodeTableName(state, nodeTableName: string | null) {
+      state.nodeTableName = nodeTableName;
+    },
```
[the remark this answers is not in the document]
gotdairyya: Closed with 63b36bc1f7f333e1304537a983526034a165a675

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,124 @@
...
+    },
+  },
+
+  methods: {
+    filterNetwork() {
+      if (this.workspace === null) {
+        return;
+      }
+
+      const aqlQuery = `FOR nodes in ${this.nodeTable} LIMIT ${this.subsetAmount} FOR v,e,p in 1..4 ANY nodes GRAPH '${store.state.networkName}' LIMIT ${this.subsetAmount} RETURN p`;
```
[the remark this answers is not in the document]
gotdairyya: Resolved the double (or more) counting with e61f2f2e77586f04f90e953d26edd128c1f80ee8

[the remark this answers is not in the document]
gotdairyya: Now outdate with new query from cd1f4153c11c4bbd4b12db06cf99e12256429297

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,123 @@
...
+        newAQLNetwork = api.aql(this.workspace, aqlQuery);
+      } catch (error) {
+        if (error.status === 400) {
+          store.commit.setLoadError({
+            message: error.statusText,
+            href: 'https://multinet.app',
+          });
+        }
+      } finally {
+        if (store.state.loadError.message === '' && typeof newAQLNetwork === 'undefined') {
```
JackWilb, line 98: `store.state.loadError.message === ''` will always equal false if there is a network error since the error is what causes this component to load. It should be `store.state.loadError.message === 'The network you are loading is too large'`. Basically if the error hasn't changed and the aql is undefined then we'd assume it was the network that isn't working.

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,123 @@
...
+
+      try {
+        newAQLNetwork = api.aql(this.workspace, aqlQuery);
+      } catch (error) {
+        if (error.status === 400) {
+          store.commit.setLoadError({
+            message: error.statusText,
+            href: 'https://multinet.app',
+          });
+        }
```
JackWilb, line 96: Add an else here as a catch all for any other type of error. It can be similar to the "Something unexpected ocurred" from `fetchNetwork`

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,123 @@
...
+      } finally {
+        if (store.state.loadError.message === '' && typeof newAQLNetwork === 'undefined') {
+          // Catches CORS errors, issues when DB/API are down, etc.
+          store.commit.setLoadError({
+            message: 'There was a network issue when getting data',
+            href: `./?workspace=${store.state.workspaceName}&graph=${store.state.networkName}`,
+          });
+        }
+      }
+      if (newAQLNetwork !== undefined) {
```
JackWilb, line 106:
```suggestion
      }

      if (newAQLNetwork !== undefined) {
```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,123 @@
...
+        newAQLNetwork.then((promise) => {
+          const aqlNetwork: Network = Object.assign(promise)[0];
```
JackWilb, line 108: Do you need the object.assign here?
```suggestion
          const aqlNetwork: Network = promise[0];
```

JackWilb (changes requested):

src/components/FilterOverlay.vue
gotdairyya: Oh great catch! Resolved with 7236c1fd9752998103169d17aeb7e37a58009e05

src/components/FilterOverlay.vue
gotdairyya: Resolved with 7236c1fd9752998103169d17aeb7e37a58009e05

src/components/Alert.vue
JackWilb: This needs to be removed

src/store/index.ts
```diff
@@ -88,7 +90,12 @@ const {
     setUserInfo(state, userInfo: UserSpec | null) {
       state.userInfo = userInfo;
     },
-
+    setNodeTableNames(state, nodeTableNames: string[]) {
+      state.nodeTableNames = nodeTableNames;
+    },
+    setEdgeTableName(state, edgeTableName: string | null) {
+      state.edgeTableName = edgeTableName;
+    },
```
JackWilb, line 98:
```suggestion

    setNodeTableNames(state, nodeTableNames: string[]) {
      state.nodeTableNames = nodeTableNames;
    },

    setEdgeTableName(state, edgeTableName: string | null) {
      state.edgeTableName = edgeTableName;
    },

```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,126 @@
...
+        </v-card>
+      </v-overlay>
+    </v-row>
+  </div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import api from '@/api';
+import { Network, Node } from '@/types';
```
JackWilb, line 60: Unused import
```suggestion
import { Network } from '@/types';
```

src/components/FilterOverlay.vue
```diff
@@ -0,0 +1,126 @@
...
+  methods: {
+    filterNetwork() {
+      if (this.workspace === null) {
+        return;
+      }
+
+      const aqlQuery = `let nodes = (FOR n in [${store.state.nodeTableNames}][**] LIMIT ${this.subsetAmount} RETURN n) let edges = (FOR e in ${store.state.edgeTableName} filter e._from in nodes[**]._id && e._to in nodes[**]._id RETURN e) 
+      RETURN {"nodes": nodes[**], edges}`;
+
+      let newAQLNetwork: Promise<any[]> | undefined;
```
JackWilb, line 83:
```suggestion
      // eslint-disable-next-line @typescript-eslint/no-explicit-any
      let newAQLNetwork: Promise<any[]> | undefined;
```

JackWilb (changes requested):
Just a couple unused imports, some whitespace, and an explicit any to fix. Then it's ready to go!

JackWilb (approved):
Great work! Thanks for being open to all my 79 requests :joy:. This is a great improvement for the app

PR #270, opened 2021-06-03
Add trrack and trrack-vis

JackWilb:
Adds trrack and trrack-vis. To keep things simple, I copied a ton of the configuration and logic from the multilink application.

I started by just tracking `selectNeighbors`, since that's a very easy way to verify that the provenance functionality is working. You can see from the vis (and the control toggle) that navigating around the trrack-vis works and that the vuex state is being updated as the provenance is updated (and vice-versa).

TODO:
- [x] Track more items in trrack (#271)


gotdairyya (approved):
Everything looks good and it is running as expected!

PR #274, opened 2021-06-11
Connectivity query

gotdairyya:
At the moment there is no functionality and the UI is ugly.

I wanted to get some initial feedback from @JackWilb about whether I set the component up correctly (I didn't use composition API).

- [x] Fix logic to show appropriate number of dropdowns (currently doubled)
- [x] Add more query logic (like _contains_ and boolean operators
- [x] Hook up AQL logic

### DIFF
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -160,8 +160,6 @@ export default {
       const queryOperator = nodeQuerySelection[1] === 'is (exact)' ? '==' : '=~';
       const aqlQuery = `let startNodes = (FOR n in [${store.state.nodeTableNames}][**] FILTER n.${nodeCategory[1]} ${queryOperator} '${nodeCategory[1]}' RETURN n) let paths = (FOR n IN startNodes FOR v, e, p IN 1..${selectedHops.value} ANY n GRAPH '${store.state.networkName}' ${pathQueryText} RETURN {nodes: p.vertices[*], edges: p.edges[*]}) let nodes = (for p in paths RETURN MERGE(p.nodes)) let edges = (for p in paths RETURN MERGE(p.edges)) RETURN {nodes: nodes, edges: edges}`;
 
-      console.log(aqlQuery);
-
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       let newAQLNetwork: Promise<any[]> | undefined;
       try {
@@ -178,7 +176,6 @@ export default {
       if (newAQLNetwork !== undefined) {
         newAQLNetwork.then((promise) => {
           const aqlNetwork: Network = promise[0];
-          console.log(aqlNetwork);
           if (aqlNetwork.nodes.length !== 0) {
           // Update state with new network
             store.commit.setNetwork(aqlNetwork);
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -175,12 +175,12 @@ export default {
     const selectedHops: Ref<number> = ref(1);
     const displayedHops = computed(() => (selectedHops.value % 2 !== 0 ? selectedHops.value + 2 : selectedHops.value + 3));
 
-    const nodeCategories = computed(() => (store.state.network ? Object.keys(store.state.network.nodes[0]) : ['No attribute selected']));
-    const nodeCategoryOptions1 = computed(() => ((store.state.network && nodeCategory1.value) ? store.state.network.nodes.map((n: Node) => n[nodeCategory1.value]).sort() : ['No network']));
-    const nodeCategoryOptions2 = computed(() => ((store.state.network && nodeCategory2.value) ? store.state.network.nodes.map((n: Node) => n[nodeCategory2.value]).sort() : ['No network']));
+    const nodeCategories = computed(() => (store.state.network ? Object.keys(store.state.network.nodes[0]) : ['No network']));
+    const nodeCategoryOptions1 = computed(() => ((store.state.network && nodeCategory1.value) ? store.state.network.nodes.map((n: Node) => n[nodeCategory1.value]).sort() : ['No attribute selected']));
+    const nodeCategoryOptions2 = computed(() => ((store.state.network && nodeCategory2.value) ? store.state.network.nodes.map((n: Node) => n[nodeCategory2.value]).sort() : ['No attribute selected']));
 
-    const edgeCategories = computed(() => (store.state.network ? Object.keys(store.state.network.edges[0]) : ['No attribute selected']));
-    const edgeCategoryOptions1 = computed(() => ((store.state.network && edgeCategory1.value) ? store.state.network.edges.map((n: Edge) => n[edgeCategory1.value]).sort() : ['No network']));
+    const edgeCategories = computed(() => (store.state.network ? Object.keys(store.state.network.edges[0]) : ['No network']));
+    const edgeCategoryOptions1 = computed(() => ((store.state.network && edgeCategory1.value) ? store.state.network.edges.map((n: Edge) => n[edgeCategory1.value]).sort() : ['No attribute selected']));
 
     function submitQuery() {
       // TODO: add ability to filter many nodes
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -60,7 +60,6 @@
 </template>
 
 <script lang="ts">
-import Vue from 'vue';
 import store from '@/store';
 import { Node, Edge } from '@/types';
 import { computed, ref, Ref } from '@vue/composition-api';
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -21,36 +21,48 @@
             v-if="i % 2 !== 0"
           >
             <v-row>
-              <v-list-item-title>
-                Node
-              </v-list-item-title>
-              <v-autocomplete
-                v-model="nodeCategory"
-                :items="nodeCategories"
-                dense
-              />
-              <v-autocomplete
-                :items="nodeCategoryOptions"
-                dense
-              />
+              <v-col>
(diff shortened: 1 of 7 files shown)


gotdairyya:
@JackWilb can you help with the following? I have stared at it too long to figure out what is wrong

- [ ] there is something strange happening where not all of the node values in the template are stored
- [ ] the modulo logic for generating the template doesn't work with odd number hops
- [ ] and I'm not sure the AQL query works (it returns more nodes than when I use the Arango platform

src/components/ConnectivityQuery.vue
```diff
@@ -0,0 +1,217 @@
...
+    });
+
+    watchEffect(() => {
+      edgeCategory.value.forEach((category: string, i: number) => {
+        // eslint-disable-next-line no-unused-expressions
+        store.state.network !== null ? edgeCategoryOptions[i] = store.state.network?.edges.map((n: Edge) => n[category]).sort() : ['No attribute selected'];
+      });
+    });
+
+    function submitQuery() {
```
JackWilb, line 147: This needs to check that all the parameters are set appropriately before firing the query. We should tell the user what needs to be addressed if it fails.

src/components/ConnectivityQuery.vue
```diff
@@ -0,0 +1,217 @@
...
+    const displayedHops = computed(() => 2 * selectedHops.value + 1);
+    const nodeQueryOptions = ['is (exact)', 'contains'];
+    const edgeQueryOptions: Ref<string[]> = ref([]);
+    const nodeCategory: Ref<string[]> = ref([]);
+    const edgeCategory: Ref<string[]> = ref([]);
+    const nodeCategorySelection: Ref<string[]> = ref([]);
+    const edgeCategorySelection: Ref<string[]> = ref([]);
+    const nodeQuerySelection: Ref<string[]> = ref([]);
+    const nodeCategories: Ref<string[][]> = ref([]);
+    const edgeCategories: Ref<string[][]> = ref([]);
```
JackWilb, line 119: These two are just the node and edge variables. I don't think you need to make these a list, with the same options repeated each time. You should just have one computed variable for all node variables, and all edge variables. Then you can reference that everywhere. I'm sure this logic exists somewhere else in this app or in the multilink app

src/components/ConnectivityQuery.vue
```diff
@@ -0,0 +1,217 @@
...
+
+export default {
+  name: 'ConnectivityQuery',
+
+  setup() {
+    const hopsSelection = [1, 2, 3, 4, 5];
+    const selectedHops: Ref<number> = ref(1);
+    const displayedHops = computed(() => 2 * selectedHops.value + 1);
+    const nodeQueryOptions = ['is (exact)', 'contains'];
+    const edgeQueryOptions: Ref<string[]> = ref([]);
```
JackWilb, line 112: Appears unused. You can remove from the return at the bottom of the `setup()` function, too
```suggestion
```

src/components/ConnectivityQuery.vue
```diff
@@ -0,0 +1,217 @@
...
+    const nodeCategory: Ref<string[]> = ref([]);
+    const edgeCategory: Ref<string[]> = ref([]);
```
JackWilb, line 114: Maybe call these `nodeVariable` and `edgeVariable`?

src/components/ConnectivityQuery.vue
```diff
@@ -0,0 +1,217 @@
...
+    const nodeCategorySelection: Ref<string[]> = ref([]);
+    const edgeCategorySelection: Ref<string[]> = ref([]);
```
JackWilb, line 116: These could then be `nodeVariableValue` and `edgeVariableValue`

src/components/ConnectivityQuery.vue
JackWilb: Then maybe call them `nodeVariableItems` and `edgeVariableItems`

src/components/ConnectivityQuery.vue
```diff
@@ -0,0 +1,217 @@
...
+
+    hopsSelection.forEach(() => {
+      nodeCategory.value.push(store.state.workspaceName === 'marclab' ? 'Label' : '');
+      edgeCategory.value.push(store.state.workspaceName === 'marclab' ? 'Type' : '');
+      nodeCategorySelection.value.push('');
+      edgeCategorySelection.value.push('');
+      nodeQuerySelection.value.push('contains');
+      nodeCategories.value.push(store.state.network ? Object.keys(store.state.network.nodes[0]) : ['No network']);
+      edgeCategories.value.push(store.state.network ? Object.keys(store.state.network.edges[0]) : ['No network']);
+    });
```
JackWilb, line 131: `hopsSelection` is only the [1,2,3,4,5]. Don't you want the `2n + 1` variable from above, `displayedHops`?

JackWilb (changes requested):
I'd like to review the AQL query, too, but the variable names and what they mean are confusing me a little. Would you mind taking care of those first (using my suggestions or coming up with better names) before I get to the query?

gotdairyya:
> I'd like to review the AQL query, too, but the variable names and what they mean are confusing me a little. Would you mind taking care of those first (using my suggestions or coming up with better names) before I get to the query?

yes totally, naming is hard, will fix!

src/components/ConnectivityQuery.vue
gotdairyya: Created issue #277

src/components/ConnectivityQuery.vue
gotdairyya: Resolved with a75f51fa8e1af5a866c69c81eb26f183ebb0f00c

src/components/ConnectivityQuery.vue
gotdairyya: Resolved with a75f51fa8e1af5a866c69c81eb26f183ebb0f00c

src/components/ConnectivityQuery.vue
gotdairyya: Resolved with a75f51fa8e1af5a866c69c81eb26f183ebb0f00c

src/components/ConnectivityQuery.vue
gotdairyya: Resolved with 7193df3411ea8b02d09f12a00a6c2d119aef0ff5

src/components/ConnectivityQuery.vue
gotdairyya: Good catch! Resolved with 803107c6eb9445bed29c8e5044705017e55d11c6

src/components/ConnectivityQuery.vue
```diff
@@ -0,0 +1,235 @@
...
+      const queryOperator = nodeQuerySelection.value[0] === 'is (exact)' ? '==' : '=~';
+      const aqlQuery = `
+        let startNodes = (FOR n in [${store.state.nodeTableNames}][**] FILTER n.${nodeVariable.value[0]} ${queryOperator} '${nodeVariableValue.value[0]}' RETURN n)
+        let paths = (FOR n IN startNodes FOR v, e, p IN 1..${selectedHops.value} ANY n GRAPH '${store.state.networkName}' ${pathQueryText} RETURN {nodes: p.vertices[*], paths: p})
+        let all_nodes = (for p in paths RETURN p.nodes)
+        let nodes_first = (for p in paths RETURN p.nodes[0])
+        let nodes_last = (for p in paths RETURN p.nodes[${selectedHops.value}])
+        let path = (for p in paths RETURN p.paths)
+        RETURN {all_nodes: UNIQUE(all_nodes[**]), nodes_first: UNIQUE(nodes_first), nodes_last: UNIQUE(nodes_last), paths: path}
+      `;
```
JackWilb, line 164: You wanted to clean this up and remove some of what is returned. We should open an issue to track that

src/components/ConnectivityQuery.vue
```diff
@@ -0,0 +1,235 @@
...
+            aqlResults.paths.forEach((path: { edges: Edge[]; vertices: Node[] }, val: number) => {
+              const newPath: Edge = {
+                _from: '', _to: '', _key: '', _id: '',
+              };
+
+              for (let i = 0; i < selectedHops.value + 1; i += 1) {
+                if (i === 0) {
+                  newPath._from = path.vertices[i]._id;
+                  if (!nodesSet.has(path.vertices[i]._key)) { newNetwork.nodes.push(path.vertices[i]); }
+                  nodesSet.add(path.vertices[i]._key);
```
JackWilb, line 196: You should use id here, I think. That would make this more robust to nodes from different tables, since keys can clash across tables

src/components/ConnectivityQuery.vue
```diff
@@ -0,0 +1,235 @@
...
+                }
+                if (i === (selectedHops.value)) {
+                  newPath._to = path.vertices[i]._id;
+                  if (!nodesSet.has(path.vertices[i]._key)) { newNetwork.nodes.push(path.vertices[i]); }
+                  nodesSet.add(path.vertices[i]._key);
```
JackWilb, line 201: Same as above

JackWilb (dismissed):
Looks great! I just saw one thing that might be worth exploring. We could probably merge without this, so I've approved

src/components/ConnectivityQuery.vue
gotdairyya: Yup, this is all part of my query rewrite in issue #278

src/components/ConnectivityQuery.vue
gotdairyya: yes ok! resolved with 48dfe4ac4154a640e38a9edcab5004a889e4e60f

src/components/ConnectivityQuery.vue
gotdairyya: Resolved with 48dfe4ac4154a640e38a9edcab5004a889e4e60f

JackWilb (approved):

Issue #278, opened 2021-06-23
ConnectivityMatrix robust querying


gotdairyya:
Update query functionality to include the following

- [ ] * for nodes
- [ ] edge filtering
- [ ] AND/OR functionality on edges
- [ ] remove unnecessary arguments
- [ ] make query not case sensitive

gotdairyya:
Closed by PR #280, didn't use the right syntax in the PR

PR #280, opened 2021-06-25
Make aql query more robust

gotdairyya:
Closes issue #278 

**To Do:**
- [x]  * for nodes + edges
- [x]  edge filtering
- [ ]  ~~AND/OR functionality on edges~~
- [x]  remove unnecessary arguments
- [x]  make query not case sensitive

### DIFF
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -148,21 +148,21 @@ export default {
         const queryOperator = nodeQuerySelection.value[i] === 'is (exact)' ? '==' : '=~';
         if (i === 0) {
           pathQueryText += `FILTER UPPER(p.vertices[${i}].${nodeVariable.value[i]}) ${queryOperator} UPPER('${nodeVariableValue.value[i]}')`;
-        } else {
+        } else if (nodeVariableValue.value[i] !== '') {
           pathQueryText += ` AND UPPER(p.vertices[${i}].${nodeVariable.value[i]}) ${queryOperator} UPPER('${nodeVariableValue.value[i]}')`;
         }
       }
       for (let i = 0; i < selectedHops.value; i += 1) {
-        if (i === 0) {
+        if (i === 0 && edgeVariableValue.value[i] !== '') {
           pathQueryText += ` FILTER p.edges[${i}].${edgeVariable.value[i]} == '${edgeVariableValue.value[i]}'`;
-        } else {
+        } else if (edgeVariableValue.value[i] !== '') {
           pathQueryText += ` AND p.edges[${i}].${edgeVariable.value[i]} == '${edgeVariableValue.value[i]}'`;
         }
       }
       const queryOperator = nodeQuerySelection.value[0] === 'is (exact)' ? '==' : '=~';
       const aqlQuery = `
         let startNodes = (FOR n in [${store.state.nodeTableNames}][**] FILTER UPPER(n.${nodeVariable.value[0]}) ${queryOperator} UPPER('${nodeVariableValue.value[0]}') RETURN n)
-        let paths = (FOR n IN startNodes FOR v, e, p IN 1..${selectedHops.value} ANY n GRAPH '${store.state.networkName}' ${pathQueryText} RETURN {nodes: p.vertices[*], paths: p})
+        let paths = (FOR n IN startNodes FOR v, e, p IN 1..${selectedHops.value} ANY n GRAPH '${store.state.networkName}' ${pathQueryText} RETURN {paths: p})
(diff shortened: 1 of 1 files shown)


gotdairyya:
Moved the AND/OR functionality to issue #282

src/components/ConnectivityQuery.vue
```diff
@@ -147,20 +147,24 @@ export default {
...
       const queryOperator = nodeQuerySelection.value[0] === 'is (exact)' ? '==' : '=~';
       const aqlQuery = `
-        let startNodes = (FOR n in [${store.state.nodeTableNames}][**] FILTER n.${nodeVariable.value[0]} ${queryOperator} '${nodeVariableValue.value[0]}' RETURN n)
-        let paths = (FOR n IN startNodes FOR v, e, p IN 1..${selectedHops.value} ANY n GRAPH '${store.state.networkName}' ${pathQueryText} RETURN {nodes: p.vertices[*], paths: p})
-        let all_nodes = (for p in paths RETURN p.nodes)
-        let nodes_first = (for p in paths RETURN p.nodes[0])
-        let nodes_last = (for p in paths RETURN p.nodes[${selectedHops.value}])
+        let startNodes = (FOR n in [${store.state.nodeTableNames}][**] FILTER UPPER(n.${nodeVariable.value[0]}) ${queryOperator} UPPER('${nodeVariableValue.value[0]}') RETURN n)
+        let paths = (FOR n IN startNodes FOR v, e, p IN 1..${selectedHops.value} ANY n GRAPH '${store.state.networkName}' ${pathQueryText} RETURN {paths: p})
         let path = (for p in paths RETURN p.paths)
```
JackWilb, line 166:
```suggestion
```

src/components/ConnectivityQuery.vue
```diff
@@ -147,20 +147,24 @@ export default {
...
-        RETURN {all_nodes: UNIQUE(all_nodes[**]), nodes_first: UNIQUE(nodes_first), nodes_last: UNIQUE(nodes_last), paths: path}
+        RETURN {paths: path}
```
JackWilb, line 167:
```suggestion
        RETURN {paths: paths[**].paths}
```

src/components/ConnectivityQuery.vue
```diff
@@ -147,20 +147,24 @@ export default {
...
+          pathQueryText += ` AND UPPER(p.vertices[${i}].${nodeVariable.value[i]}) ${queryOperator} UPPER('${nodeVariableValue.value[i]}')`;
+        }
+      }
+      for (let i = 0; i < selectedHops.value; i += 1) {
+        if (i === 0 && edgeVariableValue.value[i] !== '') {
+          pathQueryText += ` FILTER p.edges[${i}].${edgeVariable.value[i]} == '${edgeVariableValue.value[i]}'`;
+        } else if (edgeVariableValue.value[i] !== '') {
+          pathQueryText += ` AND p.edges[${i}].${edgeVariable.value[i]} == '${edgeVariableValue.value[i]}'`;
         }
       }
```
JackWilb, line 161: Why don't these have query operators?

JackWilb (changes requested):
I see one small change you can make to simplify things even more with the query

In general, I think we should offer more query operators for both nodes and edges. I can see use cases for >, <, !=, etc. Is that something you're considering?

I verified:
- [x]  * for nodes + edges
- [ ]  edge filtering
- [x]  remove unnecessary arguments
- [x]  make query not case sensitive

I can't run a query that queries the network for `edge.value == 15` with no nodes. I'm not sure if that's a query we should allow, or not, though. (looks like this is covered by #283)

src/components/ConnectivityQuery.vue
gotdairyya: Because edge selection will always be precise for the marclab... But now I'm not sure if it makes sense in other use cases

gotdairyya: > I see one small change you can make to simplify things even more with the query
> 
> In general, I think we should offer more query operators for both nodes and edges. I can see use cases for >, <, !=, etc. Is that something you're considering?
> 
> I verified:
> 
> * [x]   * for nodes + edges
> * [ ]   edge filtering
> * [x]   remove unnecessary arguments
> * [x]   make query not case sensitive
> 
> I can't run a query that queries the network for `edge.value == 15` with no nodes. I'm not sure if that's a query we should allow, or not, though. (looks like this is covered by #283)

Yes so currently the first node needs to be selected in order to run a query successfully.
I've also created an issue to address numerical queries #284 , which I think we agreed on slack will be a later functionality

JackWilb (approved):

PR #285, opened 2021-06-28
Simplify connectivity query component

JackWilb:
This change simplifies the logic of the connectivity query component by removing the arrays that existed for each specific type (nodes and edges), replacing them with one list that controls both. This change affects all instances of there being two lists performing the same function.

In so doing, this simplifies the `queryOperator` logic and allows for path filtering with the operators (currently `==` and `=~`).

Another nice thing this change gives us is a more condensed html template, since we can reuse the same arrays everywhere


gotdairyya (approved):
Looks great! Thanks for simplifying the logic + template!

PR #286, opened 2021-06-29
Move network variables to store

### LINKED ISSUE #281: ConnectivityMatrix query autofill options
the autocomplete options are computed from the network making it difficult to switch between queries

I'm considering creating store values as a solution 
@JackWilb what do you think of this as a solution?

image

gotdairyya:
Closes #281 

This change will make it possible to switch between queries without refreshing the page.
It will also show all possible options for large networks because it will pull the values from the original network.

### DIFF
--- a/src/components/FilterOverlay.vue
+++ b/src/components/FilterOverlay.vue
@@ -57,7 +57,7 @@
 <script lang="ts">
 import Vue from 'vue';
 import api from '@/api';
-import { Network } from '@/types';
+import { ArangoAttributes, Network } from '@/types';
 import store from '@/store';
 
 export default Vue.extend({
@@ -150,8 +150,8 @@ export default Vue.extend({
 
           const nodeAttrDict: {[key: string]: number} = {};
           const edgeAttrDict: {[key: string]: number} = {};
-          const nodeAttributes: {[key: string]: string[] | unknown[]} = {};
-          const edgeAttributes: {[key: string]: string[] | unknown[]} = {};
+          const nodeAttributes: ArangoAttributes = {};
+          const edgeAttributes: ArangoAttributes = {};
 
           const getKeyByValue = (obj: {[key: string]: string | number | boolean }, value: string | number | boolean) => Object.keys(obj)[Object.values(obj).indexOf(value)];
 
@@ -172,7 +172,7 @@ export default Vue.extend({
           for (const [key, value] of Object.entries(edgeAttrDict)) {
             edgeAttributes[key] = [...new Set(aqlResults.edgeValues.map((vals: string[]) => `${vals[value]}`).sort())];
           }
-          store.commit.setLargeNetworkAttributeValues(nodeAttributes, edgeAttributes);
+          store.commit.setLargeNetworkAttributeValues({ nodeAttributes, edgeAttributes });
         });
       }
     },
--- a/src/components/FilterOverlay.vue
+++ b/src/components/FilterOverlay.vue
@@ -120,6 +120,60 @@ export default Vue.extend({
             });
           }
         });
+        this.getAttributes();
+      }
+    },
+    getAttributes() {
+      const aqlQuery = `
+      let nodeValues = (FOR doc IN ${store.state.nodeTableNames}[**] RETURN VALUES(doc))
+      let edgeValues = (FOR doc IN ${store.state.edgeTableName} RETURN VALUES(doc))
+      let nodeAttr = (FOR doc IN ${store.state.nodeTableNames}[**] LIMIT 1 RETURN doc)
+      let edgeAttr = (FOR doc IN ${store.state.edgeTableName} LIMIT 1 RETURN doc)
+      RETURN {nodeAttributes: nodeAttr, nodeValues: nodeValues, edgeAttributes: edgeAttr, edgeValues: edgeValues}
+      `;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      let nodeAndEdgeQuery: Promise<any[]> | undefined;
+      try {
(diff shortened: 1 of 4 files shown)


gotdairyya:
This code mostly works, there is something going on within the FilterOverlay component where I am calculating the node and edge variables using AQL only if the network was subsetted.

The issue seems to be that the state is empty. This is the error I see: `Property 'setLargeNetworkAttributeValues' does not exist on type '{}'`

I have a feeling that it's due to async loading, but would like to discuss after you take a look

src/components/FilterOverlay.vue
```diff
@@ -120,6 +120,60 @@ export default Vue.extend({
             });
           }
         });
+        this.getAttributes();
+      }
+    },
+    getAttributes() {
```
JackWilb, line 126:
```suggestion
    },

    getAttributes() {
```

src/components/FilterOverlay.vue
```diff
@@ -120,6 +120,60 @@ export default Vue.extend({
...
+      const aqlQuery = `
+      let nodeValues = (FOR doc IN ${store.state.nodeTableNames}[**] RETURN VALUES(doc))
+      let edgeValues = (FOR doc IN ${store.state.edgeTableName} RETURN VALUES(doc))
+      let nodeAttr = (FOR doc IN ${store.state.nodeTableNames}[**] LIMIT 1 RETURN doc)
+      let edgeAttr = (FOR doc IN ${store.state.edgeTableName} LIMIT 1 RETURN doc)
+      RETURN {nodeAttributes: nodeAttr, nodeValues: nodeValues, edgeAttributes: edgeAttr, edgeValues: edgeValues}
+      `;
```
JackWilb, line 133: From the [arango docs](https://www.arangodb.com/docs/stable/aql/fundamentals-document-data.html):

>Data in collections is stored in documents, with each document potentially having different attributes than other documents. This is true even for documents of the same collection.

Therefore, this might not get us all key value pairs. We might need to reformat this query a little

JackWilb (changes requested):
I think there are some edge cases that need to be considered, but overall, this works great!

JackWilb:
Also, I didn't test this locally. I think we should test an example of 1 node table and 2 or more node tables to make sure that expansion works

src/types.ts
```diff
@@ -66,6 +66,8 @@ export interface State {
   edgeTableName: string | null;
   provenance: Provenance<State, ProvenanceEventTypes, unknown> | null;
   showProvenanceVis: boolean;
+  nodeAttributes: {[key: string]: string[]};
+  edgeAttributes: {[key: string]: string[]};
```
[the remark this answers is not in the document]
gotdairyya: If we are setting types for everything on upload (hypothetically) I assume we will see the types from upload options -- or am I misinterpreting that? right now everything I have seen is stored as a string

src/store/index.ts
```diff
@@ -234,6 +236,25 @@ const {
...
+      state.edgeAttributes = edgeKeys.reduce((ac, a) => ({ ...ac, [a]: [] }), {});
+      edgeKeys.forEach((key: string) => {
+        state.edgeAttributes[key] = [...new Set(network.edges.map((e: Edge) => `${e[key]}`).sort())];
+      });
+    },
+
+    setLargeNetworkAttributeValues(state: State, nodeAttributes: {[key: string]: string[]}, edgeAttributes: {[key: string]: string[]}) {
+      state.nodeAttributes = nodeAttributes;
+      state.edgeAttributes = edgeAttributes;
+    },
```
JackWilb, line 257:
```suggestion
    setLargeNetworkAttributeValues(state: State, payload: { nodeAttributes: { [key: string]: string[] }; edgeAttributes: { [key: string]: string[] }}) {
      state.nodeAttributes = payload.nodeAttributes;
      state.edgeAttributes = payload.edgeAttributes;
    },
```

src/types.ts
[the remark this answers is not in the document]
JackWilb: I made this fix. I just chose unknown since we don't know which it will be exactly

src/store/index.ts
JackWilb: See 00e4895

src/components/FilterOverlay.vue
gotdairyya: I do think it gets all key-value pairs...We should definitely test this though, so I opened issue #287 for this (mostly because I don't trust AQL)

src/store/index.ts
```diff
@@ -234,6 +236,25 @@ const {
...
+      nodeKeys.forEach((key: string) => {
+        state.nodeAttributes[key] = [...new Set(network.nodes.map((n: Node) => `${n[key]}`).sort())];
+      });
+
+      const edgeKeys = Object.keys(network.edges[0]);
+      state.edgeAttributes = edgeKeys.reduce((ac, a) => ({ ...ac, [a]: [] }), {});
+      edgeKeys.forEach((key: string) => {
+        state.edgeAttributes[key] = [...new Set(network.edges.map((e: Edge) => `${e[key]}`).sort())];
+      });
+    },
```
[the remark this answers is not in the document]
gotdairyya: Closed with 365f9591157159919a5bdfbc77abc63f2b741143

gotdairyya:
> Also, I didn't test this locally. I think we should test an example of 1 node table and 2 or more node tables to make sure that expansion works

I tested with the Paul Revere dataset. It is not pulling from the second table. I will figure out why.

Also, I am getting a bug with nodes > 1, and edges > 1 in the dropdown. 

To Do:
- [ ] Get values for all node tables
- [ ] Address issue with dropdown logic

JackWilb (approved):

Issue #224, opened 2021-03-22
Sorting behavior is strange when network is aggregated


JackWilb:
From `pranav-aggr`

JackWilb:
This is better described in #294 and #190

PR #288, opened 2021-07-07
Refactor aggregation logic and allow it to be applied to the matrix again

### LINKED ISSUE #118: Keep original data from Multinet in a data prop
This will allow us to use a *computed* prop to derive the D3-formatted data that's needed for rendering, without having to write two sets of processing routines for things like creating supernode networks, etc.

In particular, this will allow us to merge the `defineNeighbors()` and `defineSuperNeighbors()` into a single function.

See https://github.com/multinet-app/multimatrix/pull/110#discussion_r510285046 for original context.
Also https://github.com/multinet-app/multimatrix/pull/110/files#r510336723 for another relevant comment.

### LINKED ISSUE #137: Fix Origin for Supernode Aggregation
Address this request https://github.com/multinet-app/multimatrix/pull/136#pullrequestreview-546652172

### LINKED ISSUE #225: Row attributes do not get reset when aggregation is disabled

### LINKED ISSUE #232: Triggering aggregation more than once breaks things

### LINKED ISSUE #238: Enable eslint rules in aggregation

JackWilb:
Closes #238
Closes #232
Closes #225
Relates to #224
Closes #137
Closes #118 (since we've refactored the whole thing)

TODOs:
- [x] Allow matrix to be reset
- [x] Show child legend
- [x] Debug issue when turning off aggregation more than once


gotdairyya:
Found one major bug (though seems like it is an easy fix) + a couple of minor ones (that are more opinions than required fixes). Otherwise, everything is working as expected!

**Major bug**
If you submit a connectivity query, view it using the aggregation, then submit another query, enable aggregation is not reset.  I _think_ it should be as simple as adding `store.dispatch.updateEnableAggregation(false)` to line 189 in `ConnectivityQuery.vue`

**Minor Bugs**
(1) If you hover over the + icon, the tooltip shows up very far away
image

(2) I would assume if you aggregate by a categorical variable you would see it as an attribute on the parent node. In the example below, the Label attribute shows up as `-` for the aggregated row. But I don't think it's a big deal, it just means you can't filter based on parent node attributes .. 
image

(3) Can we update the child class text color to be slate gray so that it is more visually distinct?

JackWilb:
Thanks for such thorough testing.

> **Major bug**
> If you submit a connectivity query, view it using the aggregation, then submit another query, enable aggregation is not reset. I _think_ it should be as simple as adding `store.dispatch.updateEnableAggregation(false)` to line 189 in `ConnectivityQuery.vue`

Your suggested fix seemed to work here. 8a5785f

> **Minor Bugs**
> (1) If you hover over the + icon, the tooltip shows up very far away
> image

I think this is a general bug in the rendering of the tooltip. I'll address it in a follow up #289 

> (2) I would assume if you aggregate by a categorical variable you would see it as an attribute on the parent node. In the example below, the Label attribute shows up as `-` for the aggregated row. But I don't think it's a big deal, it just means you can't filter based on parent node attributes ..
> image

Addressed this in 0f21621.

> (3) Can we update the child class text color to be slate gray so that it is more visually distinct?

Fixed in 31fe5e7

> (4) Color scale of child legend doesn't match the child cells in the matrix

I'll fix in #290

> (5) Th z-index of lineup is higher than the tooltips, so sometimes it blocks the tooltip.

I'll add this to #289.

gotdairyya (approved):
Looks great! Thanks Jack for addressing my concerns!

PR #292, opened 2021-07-09
Update tooltip logic so that position is consistent + show cell message

### LINKED ISSUE #289: Fix tooltip location when hovering and z-index
This doesn't seem to be fixed, especially so on the foreign objects for an aggregated network. See https://github.com/multinet-app/multimatrix/pull/288#issuecomment-875849035

JackWilb:
Closes #289

I fixed the logic that sets the location for the tooltip, reordered the events so that the cell event is not overwritten by a row event, and added a z-index so it doesn't get hidden by lineup. The end result is that the location is now consistent, the cell hovering now shows the right message, and the tooltip doesn't get hidden by lineup elements.


gotdairyya (approved):
Looks grrrreat!

JackWilb:
🐯

PR #293, opened 2021-07-09
Remove all child node logic, child nodes are now "normal" nodes

### LINKED ISSUE #290: Fix child legend color scale colors
It's slightly off from the actual color of the nodes for some reason

JackWilb:
Closes #290

I chose to use one scale for all "normal" nodes and one for super nodes. This means that lots of the child node logic can be removed and simplifies the code quite a bit. 

To keep the child legend, I used a ternary operatore to show/hide the legend and to change the name of the legend.

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -67,10 +67,6 @@ export default Vue.extend({
       return store.state.aggregated;
     },
 
-    showChildLegend() {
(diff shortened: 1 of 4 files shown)


gotdairyya (approved):
Yup! A definite improvement to remove one level of complexity in naming!

PR #300, opened 2021-07-12
Alleviate grid illusion by removing cell radius and darkening lines

### LINKED ISSUE #92: Alleviate grid illusion effect in clusters of squares
When a cluster of cells is filled in, we get a [grid illusion](https://en.wikipedia.org/wiki/Grid_illusion) effect that is rather distracting/annoying. It's made worse by the fact that the grid lines cross at precisely these points, lending some substance on which the illusion might be building.

Recommendations from Jared:
- [ ] reduce border-radius value
- [ ] slightly darken the grid lines

JackWilb:
Closes #92

Using Roni's/Jared's suggestions and [this article](https://ux.stackexchange.com/questions/53791/avoiding-the-hermann-grid-illusion) I've attempted to fix this problem.

I removed the border radius on the cells and darkened the grid lines. It seems to remove the effect for me, but I'd like a quick sanity check that it works on other machines.

### DIFF
--- a/src/components/MultiMatrix.vue
+++ b/src/components/MultiMatrix.vue
@@ -363,9 +363,6 @@ export default Vue.extend({
 
       this.processData();
 
-      // set the radius for cells
-      const cellRadius = 3;
-
       // set the matrix highlight
       const matrixHighlightLength = this.matrix.length * this.cellSize;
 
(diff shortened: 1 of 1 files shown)


gotdairyya:
It looks like the right and bottom outer border is being cut-off, other than that looks great!

gotdairyya (approved):
looks great! (though personally I like the cellRadius >_<)

PR #301, opened 2021-07-12
Remove internal/arango vars

### LINKED ISSUE #241: Remove non-attribute attributes
Filter out `_id`, `_key`,` _rev` from node attributes on network load.

@JackWilb what do you think?

JackWilb:
Closes #241

I adopted the logic from MultiLink to filter out internal variables from the list of node variables and link variables. Since the node variables are tracked in the store, it was a simple fix. The link variables also only appear in one spot


gotdairyya (approved):
Yeah, awesome addition!

PR #302, opened 2021-07-13
Draw intermediary nodes matrix

gotdairyya:
This table should show the middle paths of a connectivity query.

To do:

- [x] add labels for rows
- [x] add symbols above columns to indicate path position


gotdairyya:
@JackWilb this is a super simple implementation of the intermediary node view

I've added issues for the next steps:
#303 
#304

src/components/IntermediaryNodes.vue
```diff
@@ -0,0 +1,246 @@
...
+    function buildIntView() {
+      if (matrix.length > 0) {
+        const svg = select('#intNode').append('g').attr('transform', `translate(${margin.left},${margin.top})`);
+
+        //   Draw path symbols
+        const circles = svg.selectAll('g.circles')
+          .data([...Array(pathLength.value).keys()])
+          .enter()
+          .append('g')
+          .attr('transform', `translate(${matrixWidth.value * 0.13}, -${margin.top / 4})`);
```
JackWilb, line 117: This numbers seem kind of random. Is there a good reason why it's `0.13` and `... / 4`? Could we make that more obvious for other people who read this code?

src/components/IntermediaryNodes.vue
```diff
@@ -0,0 +1,246 @@
...
+          .text((_, i: number) => sortOrder.value[i]);
+
+        //   Draw columns
+        svg
+          .selectAll('g.column')
+          .data(matrix)
+          .enter()
+          .append('g')
+          .attr('class', 'column')
+          .attr('transform', (_, i: number) => `translate(${xScale(i)}, 0)rotate(-90)`);
```
JackWilb: Is this code a remnant from something else? Maybe we can remove it :)

JackWilb (commented):
It seems like if you run a 2 hop query and then a 3 hop, it doesn't update properly. Maybe that's a more general problem here, that it's not reactive to new update. We'd need to explore that a bit more

Other than that, I've cleaned up some type issues, etc. I also have a couple of questions  for you

src/components/IntermediaryNodes.vue
```diff
@@ -0,0 +1,246 @@
...
+    });
+
+    watchEffect(() => {
+      matrixHeight.value = connectivityPaths.value.nodes.length > 0 ? connectivityPaths.value.nodes.length * cellSize.value + margin.top + margin.bottom : 0;
+    });
+
+    const intNodeWidth = computed(() => (store.state.connectivityMatrixPaths.nodes.length > 0
+      ? store.state.connectivityMatrixPaths.nodes.length * cellSize.value + margin.left + margin.right
+      : 0));
+    const sortOrder = computed(() => store.state.connectivityMatrixPaths.nodes.map((node) => node._key).sort());
```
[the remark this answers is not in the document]
gotdairyya: No I don't think so. I'm checked that I'm using sortOrder where the order matters, but definitely worth discussing a little

src/components/IntermediaryNodes.vue
gotdairyya: yes. this is what I mentioned earlier about the scrappy bits of the code. will try to clean up with more reasonable variables

src/components/IntermediaryNodes.vue
gotdairyya: Yes! good catch!!

src/components/MultiMatrix.vue
```diff
@@ -169,6 +171,10 @@ export default Vue.extend({
     parentColorScale() {
       return store.getters.parentColorScale;
     },
+
+    showInterMediateNodes() {
```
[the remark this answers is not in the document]
gotdairyya: Resolved with 5a6611b029d830a52dac77fa659b6a3ac85f277f

src/components/IntermediaryNodes.vue
```diff
@@ -0,0 +1,246 @@
...
+    const matrixWidth: Ref<number> = ref(0);
+    const matrixHeight: Ref<number> = ref(0);
+
+    watchEffect(() => {
+      matrixWidth.value = connectivityPaths.value.nodes.length > 0 ? connectivityPaths.value.paths[0].edges.length * cellSize.value + margin.left + margin.right : 0;
+    });
+
+    watchEffect(() => {
+      matrixHeight.value = connectivityPaths.value.nodes.length > 0 ? connectivityPaths.value.nodes.length * cellSize.value + margin.top + margin.bottom : 0;
+    });
```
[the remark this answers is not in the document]
gotdairyya: Yeah still figuring out when to use watcheffect v. computed
Resolved with 5a6611b029d830a52dac77fa659b6a3ac85f277f

src/components/IntermediaryNodes.vue
gotdairyya: Resolved with 5a6611b029d830a52dac77fa659b6a3ac85f277f

src/components/IntermediaryNodes.vue
gotdairyya: Resolved with b099d20fdc33025a36e61fd774a7df1471277af1

src/components/IntermediaryNodes.vue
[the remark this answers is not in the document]
JackWilb: I see that the labels are right next to the vis, it's probably not an issue

src/components/IntermediaryNodes.vue
```diff
@@ -0,0 +1,237 @@
...
+    id="intNodeDiv"
+    :style="`width: ${intNodeWidth}px;`"
+  >
+    <svg
+      id="intNode"
+      :width="matrixWidth"
+      :height="matrixHeight"
+      :viewbox="`0 0 ${matrixWidth} ${matrixHeight}`"
+    />
+  </div>
```
JackWilb, line 216:
```suggestion
  <svg
    id="intNode"
    :width="matrixWidth"
    :height="matrixHeight"
    :viewbox="`0 0 ${matrixWidth} ${matrixHeight}`"
  />
```

JackWilb (changes requested):
There's a bunch of extra space that is added if this div is present. You could also open an issue to address the reactivity of the component since it doesn't seem to update if the path length changes or a query is re-run

With div:
image

Without div:
image

src/components/IntermediaryNodes.vue
gotdairyya: I have the div in there to add the extra padding. Do you think I need to remove it?

gotdairyya:
Ok so aside from the matrix not updating correctly (pushed to issue #311) I think I have addressed all of the issues
Let me know if I should do anything else before merge?

src/components/IntermediaryNodes.vue
JackWilb: I thought it might have a bit too much, but it's okay. We can ask Jared for some feedback at standup

JackWilb (approved):

PR #307, opened 2021-07-13
Make query pretty

gotdairyya:
Focused on readability and some UX
Super open to suggestions!

### DIFF
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -71,6 +71,7 @@
           class="ml-0 mt-4"
           color="primary"
           depressed
+          :loading="loading"
           @click="submitQuery"
         >
           Submit Query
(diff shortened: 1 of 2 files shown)


JackWilb:
My feedback so far is that it'd be nice if the nodes/edges were numbered (e.g. start node, int node 1, end node, etc.).

We could also probably use some icons, but I'm not sure which and where to place them for them to be most effective.

jtomeck (approved):
I committed some changes just now. Let me know what you think.

MDI comes with icons inside circles up to 10. I think that should be enough, but do you forsee the number of hops growing at all? I think that could be a potential issue with what I've done.

JackWilb (approved):
I like these changes, Derya and Jared. The query shouldn't grow over 5 hops, that's already probably too much and it's very intensive to query that far. This design should work into the future!

gotdairyya:
Yes, this looks great, thanks, Jared and Jack. I'll merge master into this branch and then if one of you can approve the PR I will go ahead and merge it

jtomeck (approved):
💯

PR #308, opened 2021-07-14
Replace aggregation UI with a single v-autocomplete

### LINKED ISSUE #291: Turn aggregation switch to drop down with default = 'none'

JackWilb:
Closes #291

Replace the enable aggregation switch and drop down with a single v-autocomplete that handles it all. The default is `'none'`. This provides a slightly better UX since a user can now switch between variables without disabling the aggregation first, and there are less click in general.

### DIFF
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -187,7 +187,7 @@ export default {
             });
 
             // Update state with new network
-            store.dispatch.updateEnableAggregation(false);
+            store.dispatch.aggregateNetwork('none');
             store.dispatch.updateNetwork({ network: newNetwork });
           }
         });
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -228,6 +228,18 @@ export default Vue.extend({
             <v-list-item-content> Directional Edges </v-list-item-content>
           </v-list-item>
 
+          <!-- Connectivity Query List Item -->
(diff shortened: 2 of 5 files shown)


gotdairyya (changes requested):
Great change! After a query is submitted, could you have the dropdown option change back to none? Otherwise, looks great!

JackWilb:
> Great change! After a query is submitted, could you have the dropdown option change back to none? Otherwise, looks great!

Don't you think it should show what we're currently aggregating by?

gotdairyya:
> > Great change! After a query is submitted, could you have the dropdown option change back to none? Otherwise, looks great!
> 
> Don't you think it should show what we're currently aggregating by?

I think that if they are submitting a hop query there are too many variables and my understanding is that the aggregation is reset when a new network is visualized

JackWilb:
> > > Great change! After a query is submitted, could you have the dropdown option change back to none? Otherwise, looks great!
> > 
> > 
> > Don't you think it should show what we're currently aggregating by?
> 
> I think that if they are submitting a hop query there are too many variables and my understanding is that the aggregation is reset when a new network is visualized

Oh, I see. I was just talking about in the drop down in general. You mean when we do a connectivity query, right? I'll see what I can do

JackWilb:
@gotdairyya see 04334b8

gotdairyya (approved):
Thank you for addressing my concern!

PR #310, opened 2021-07-14
Convert all components to composition API

### LINKED ISSUE #247: Convert FilterOverlay into composition API
Change FilterOverlay so that it follows the composition API style

JackWilb:
Depends on #308 ~~(will need one additional change in controls.vue since this branch missing a commit that adds a watcher)~~
Closes #247

Converts the remaining options API over to composition. It didn't require too much work, but we could improve this by grouping together relevant code so it's not so spread out, especially in MultiMatrix.vue

### DIFF
--- a/src/components/AboutDialog.vue
+++ b/src/components/AboutDialog.vue
@@ -54,14 +54,16 @@
 </template>
 
 <script lang="ts">
-import Vue from 'vue';
+import { ref } from '@vue/composition-api';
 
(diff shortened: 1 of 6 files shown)


gotdairyya (approved):
Wow yeah. That seemed like a lot of work. The code is much more readable now though, thank you!

JackWilb:
Force pushed to fix the problem next to "depends on #308" in the description.

PR #312, opened 2021-07-15
Use defineComponent in each component

JackWilb:
This is an improvement that I found doing some work on MultiLink (with the help of Roni and Jake). We need to use `defineComponent` when using props (which we're not), but it's probably just best to use it all the time. This means that if we want to use a new feature, we'll likely get the typescript support through this function


JackWilb:
Skipping review since this is a minor change

PR #313, opened 2021-07-15
Fix the selected cells watcher

JackWilb:
After my recent changes to the composition API, I noticed that the cells weren't able to be selected. This code fixes that issue


JackWilb:
Small change, merging without review

PR #316, opened 2021-07-19
Display path table

gotdairyya:
There are a few things left to do:

- [x] Fix logic for selecting/deselecting a cell
- [x] Find a better placement for the table (@JackWilb would love your opinions)
- [x] Make it possible so that the user can select what values they see for each node/edge along the path

### DIFF
--- a/src/components/IntermediaryNodes.vue
+++ b/src/components/IntermediaryNodes.vue
@@ -1,6 +1,6 @@
 <script lang="ts">
 import {
-  computed, onMounted, watch,
+  computed, onMounted, ref, watch,
 } from '@vue/composition-api';
 import {
   scaleLinear,
@@ -8,9 +8,14 @@ import {
 import { select } from 'd3-selection';
 import store from '@/store';
 import { ConnectivityCell } from '@/types';
+import PathTable from '@/components/PathTable.vue';
 
 export default {
   name: 'IntermediaryNodes',
+
+  components: {
+    PathTable,
+  },
 
   setup() {
     const network = computed(() => store.state.network);
@@ -19,6 +24,7 @@ export default {
     const cellSize = computed(() => store.state.cellSize);
     const pathLength = computed(() => connectivityPaths.value.paths[0].vertices.length);
     const edgeLength = computed(() => connectivityPaths.value.paths[0].edges.length);
+    const cellSelected = ref(false);
 
     const margin = {
       top: 79,
@@ -95,7 +101,10 @@ export default {
         .style('fill-opacity', (d) => opacity(d.z))
         .style('fill', 'blue');
 
-      cell.on('click', () => console.log('clicked!', rowData));
+      cell.on('click', () => {
+        cellSelected.value = !cellSelected.value;
+        store.commit.setSelectedConnectivityPaths(rowData);
+      });
 
       cell.append('title').text((d) => `${d.cellName} in ${d.z} paths`);
     }
@@ -196,6 +205,7 @@ export default {
     });
 
     return {
+      cellSelected,
       intNodeWidth,
       matrixWidth,
       matrixHeight,
@@ -215,6 +225,7 @@ export default {
       :height="matrixHeight"
       :viewbox="`0 0 ${matrixWidth} ${matrixHeight}`"
     />
+    <path-table v-if="cellSelected" />
   </div>
 </template>
 
--- a/src/components/IntermediaryNodes.vue
+++ b/src/components/IntermediaryNodes.vue
@@ -1,7 +1,6 @@
 <script lang="ts">
 import {
-  computed, onMounted, watch,
-  defineComponent,
+  computed, defineComponent, onMounted, watch,
 } from '@vue/composition-api';
 import {
   scaleLinear,
@@ -226,7 +225,7 @@ export default defineComponent({
 <template>
   <div
     id="intNodeDiv"
-    :style="`width: 700px;`"
+    :style="`width: ${intNodeWidth}px;`"
   >
     <svg
       id="intNode"
--- a/src/components/IntermediaryNodes.vue
+++ b/src/components/IntermediaryNodes.vue
@@ -1,6 +1,6 @@
 <script lang="ts">
 import {
-  computed, onMounted, ref, watch,
+  computed, onMounted, watch,
   defineComponent,
 } from '@vue/composition-api';
 import {
@@ -9,14 +9,9 @@ import {
 import { select } from 'd3-selection';
 import store from '@/store';
 import { ConnectivityCell } from '@/types';
-import PathTable from '@/components/PathTable.vue';
 
 export default defineComponent({
   name: 'IntermediaryNodes',
-
-  components: {
-    PathTable,
-  },
 
   setup() {
     const network = computed(() => store.state.network);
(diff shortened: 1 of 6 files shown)


JackWilb:
This doesn't seem to be working for me. Here's what I see: 
image

I can't see the rows

src/components/IntermediaryNodes.vue
```diff
@@ -1,6 +1,7 @@
 <script lang="ts">
 import {
-  computed, defineComponent, onMounted, watch,
+  computed, onMounted, watch,
+  defineComponent,
```
JackWilb, line 4: Let's revert this change

src/components/IntermediaryNodes.vue
```diff
@@ -87,13 +97,24 @@ export default defineComponent({
...
 
+      cell.on('click', () => {
+        if (selectedCell !== rowData[0].cellName || !showTable.value) {
+          selectedCell = rowData[0].cellName;
+          store.commit.setSelectedConnectivityPaths(rowData);
+          showTable.value = true;
+        } else if (selectedCell === rowData[0].cellName && showTable.value) {
+          showTable.value = !showTable.value;
+        }
+      });
```
JackWilb, line 116: Maybe we can set a class here that shows which cell is highlighted? Just some more visual feedback so you can see which is active

src/components/IntermediaryNodes.vue
```diff
@@ -204,7 +226,7 @@ export default defineComponent({
 <template>
   <div
     id="intNodeDiv"
-    :style="`width: ${intNodeWidth}px;`"
+    :style="`width: 700px;`"
```
JackWilb, line 229: I'm curious about this change. Why 700? and if we go with 700, we need to remove all references to `intNodeWidth` in `<script>`

src/components/IntermediaryNodes.vue
```diff
@@ -233,4 +255,10 @@ svg >>> .rowLabels {
   font-size: 12pt;
   z-index: 100;
 }
+
+svg >>> .connectivityCell:hover {
+  cursor: pointer;
+  stroke-width: 1px;
+  stroke: black;
+}
```
JackWilb, line 263: We could do something similar here for the selection as I mentioned above

src/components/PathTable.vue
```diff
@@ -0,0 +1,199 @@
...
+        <v-card-title class="pt-0 mt-0">
+          <v-text-field
+            v-model="search"
+            append-icon="mdi-magnify"
+            label="Search"
+            single-line
+            hide-details
+          />
+        </v-card-title>
+      </v-card>
```
JackWilb, line 31: Do these need to be in v-card wrappers? Can't we just have the text-field and style it?

src/components/PathTable.vue
```diff
@@ -0,0 +1,199 @@
...
+              <v-row>
+                <v-col
+                  v-for="(header, i) in headers"
+                  :key="i"
+                  class="pa-0"
+                >
+                  <v-select
+                    v-model="selectedHeader[i]"
+                    :items="i % 2 ? headerEdgeSelections : headerNodeSelections"
+                    label="Attribute"
```
JackWilb, line 51:
```suggestion
                    label="i % 2 ? 'Edge Attribute' : 'Node Attribute'"
```

src/components/PathTable.vue
```diff
@@ -0,0 +1,199 @@
...
+        <template
+          v-slot:top
+        >
+          <v-list dense>
+            <v-list-item dense>
+              <v-row>
+                <v-col
+                  v-for="(header, i) in headers"
+                  :key="i"
+                  class="pa-0"
```
JackWilb, line 46: I think this needs some x padding
```suggestion
                  class="py-0"
```

src/components/PathTable.vue
```diff
@@ -0,0 +1,199 @@
...
+        :items="tableData"
+        :search="search"
+      >
+        <template
+          v-slot:top
+        >
+          <v-list dense>
+            <v-list-item dense>
+              <v-row>
+                <v-col
```
JackWilb, line 43: These v-col alignments get messed up when changing data. Can we set a constant width on the table columns so they always line up?

src/components/PathTable.vue
```diff
@@ -0,0 +1,199 @@
...
+
+export default {
+  name: 'PathTable',
+
+  setup() {
+    const search = ref('');
+    const pathLength = computed(() => store.state.selectedConnectivityPaths[0].vertices.length);
+
+    const headerNodeSelections = computed(() => store.getters.nodeVariableItems);
+    const headerEdgeSelections = (store.state.network ? Object.keys(store.state.edgeAttributes).filter((varName) => !isInternalField(varName)) : ['No network']);
```
JackWilb, line 92: Is this used elsewhere in the codebase? Maybe we should integrate it into the store?

src/components/PathTable.vue
JackWilb: Alternatively, you could use the `headers` list and append attribute to it like `headers[i] ${Attribute}`

JackWilb (commented):
I think this PR is 95% ready to go. Thanks for your hard work to get it to this state! I had a couple of code suggestions that might clean things up a bit, but I don't think there needs to be any big changes

src/components/IntermediaryNodes.vue
gotdairyya: that was an old thing that I forgot to switch back to

src/components/IntermediaryNodes.vue
gotdairyya: Closed with f5202f02b539ffd5cef3c921644ab4794201711a

src/components/IntermediaryNodes.vue
gotdairyya: Closed with f5202f02b539ffd5cef3c921644ab4794201711a

src/components/IntermediaryNodes.vue
gotdairyya: good suggestion! closed with bc088054db4be78c96ef84bd7eb00e3c141ccd5b

src/components/IntermediaryNodes.vue
gotdairyya: closed with bc088054db4be78c96ef84bd7eb00e3c141ccd5b

src/components/PathTable.vue
gotdairyya: so I removed the card, but I styled it a little differently, let me know what you think!

src/components/PathTable.vue
gotdairyya: Addressed with cd98d15584e8c87ac42ca71cc20f1ec858aedf8a

src/components/PathTable.vue
gotdairyya: Addressed with cd98d15584e8c87ac42ca71cc20f1ec858aedf8a

JackWilb: I think it's okay, but it might be better with the search on the next line down. I don't think it needs to fill the horizontal space there

src/components/PathTable.vue
gotdairyya: should be fixed with db1f89f7ad526e99e79b035423e2f430196a8b61

src/components/PathTable.vue
gotdairyya: Good suggestion!

src/components/PathTable.vue
gotdairyya: I don't think the search bar needs that much space though

src/components/PathTable.vue
gotdairyya: Yes it was, added it to store c80e3517e73ddec5b3d8e7e4f3f7ccde54b7a97e

src/components/PathTable.vue
JackWilb: Woo! Perfect!

src/components/PathTable.vue
JackWilb: Okay

JackWilb (approved):
I just updated the location that the card mounts at and fixed all the type issues 🤞🏻 Hopefully that allows it to build. If you're happy with my changes (and they look good on your machine, too) go ahead and merge 🎉

PR #317, opened 2021-07-27
Add function to throw out old visual before re-rendering int node vis

### LINKED ISSUE #311: IntNode: Fix update
Currently on update the int node matrix does not redraw nicely

JackWilb:
Closes #311

Removing the visual elements and redrawing is the fastest path to what we want. We should check that we're not missing any data because of a weird enter/exit issue.

Let me know what you think @gotdairyya. We could also try to approach this in a smarter way, using ids and the proper d3 enter exit merge, but it might be more work than it's worth

### DIFF
--- a/src/components/IntermediaryNodes.vue
+++ b/src/components/IntermediaryNodes.vue
@@ -130,8 +130,6 @@ export default defineComponent({
     function buildIntView() {
       if (matrix.length > 0) {
         const svg = select('#intNode').append('g').attr('transform', `translate(${margin.left},${margin.top})`);
-        const svgWidth: number = parseFloat(select('#intNode').attr('width'));
-        const svgHeight: number = parseFloat(select('#intNode').attr('height'));
         const rowLabelWidth = 20;
 
         //   Draw path symbols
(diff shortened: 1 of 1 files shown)


gotdairyya:
Not sure why your code doesn't work. There are 2 main issues

(1) it does not update the gridlines:

_This is switching from 2 hops to 3 hops_
image

(2) The color scale doesn't reset
_This is switching from 3 hops to 2 hop_
image

gotdairyya (changes requested):
It looks like it should work, but the elements aren't being rebuilt correctly when switching across different hop size queries

gotdairyya:
those edits fixed most of the issues! 

two things coming up though:
(1) I'm still not seeing the correct color scale though when I go from 3 hop to 2 hop

(2) in going from a 2 hop query to a 1 hop query (so the int node table disappears) and then back to a 2 hop this happens:
image

gotdairyya (changes requested):
The gridlines are redrawn correctly, but the color scales don't update

JackWilb:
Looks good with 3503454 and fc27a71

gotdairyya (approved):

PR #320, opened 2021-08-05
Add cellSize slider and logic to handle resize

JackWilb:
This adds ability to change cell size from the control panel and resizes all the relevant pieces of the matrix.

I also took the opportunity to clean up some of the parameters here and just make it a bit easier to deal with.

This doesn't work for the lineup view, yet. We should open an issue to deal with that.

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -46,6 +46,14 @@ export default defineComponent({
         store.commit.setShowGridlines(value);
       },
     });
+    const cellSize = computed({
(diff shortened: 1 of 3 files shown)


gotdairyya (approved):
Two things that can be pushed to issues:
1. The label sizes don't change proportionately 
2. The int node table should also be updated to be reactive
image

JackWilb:
Opened #323, #324, #325 to track the remnants

PR #321, opened 2021-08-05
Make the linter fail on warnings

JackWilb:
This will make the `yarn lint` command function the same way it does on multilink and the client. 

A byproduct of this change is that it will be harder to commit code that doens't pass the linter's validation. This could cause some more work when coding, but it will mean that it won't need to be tidied up on PR review.

You can always bypass this with `git commit -m "message" --no-verify` if we're going to defer the work until the review is over


gotdairyya (approved):
Looks good!

PR #322, opened 2021-08-05
Add dynamic columns to table

gotdairyya:
Instead on choosing only one column header, users can now pick multiple column headers in the path table


JackWilb (approved):
Appears to work as intended! I have a couple of suggestions that can become issues (or not):

- Is there a way to show which attributes belong to each node, either with header spacing, or with some other kind of visual indicator?
- The attributes get reset on close and open. Maybe we can have them persist?

PR #328, opened 2021-08-30
Make int nodes table reactive to cell size

### LINKED ISSUE #324: Make int nodes vis update with cellSize
Remnant from #320

JackWilb:
Closes #324

I refactored some of this logic to work with the changing cellSize. In so doing, I've simplified the logic slightly using translates where possible to remove the number of positioning steps. 

The new set up is 1 cell width for labels, 0.5 cell width of padding either side of the cells and then 1 cell width for each row of cells. This seems to work pretty well at the moment. We'll need to do some testing to make sure this generalizes well

I haven't been able to test this with the full size since the matrix becomes too large for my screen

New behavior for small cell size:
image

New behavior for large cell size:
image

TODO:
- [x] Some testing
- [x] The labels appear to get occluded when the cell size gets large enough. There must be some positioning causing problems

### DIFF
--- a/src/components/IntermediaryNodes.vue
+++ b/src/components/IntermediaryNodes.vue
@@ -127,7 +127,6 @@ export default defineComponent({
 
     function buildIntView() {
       if (matrix.length > 0) {
-        const labelPadding = 10;
         const headerPadding = 5;
         const circleRadius = cellSize.value / 2;
         const cellFontSize = cellSize.value * 0.8;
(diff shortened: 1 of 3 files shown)


JackWilb:
The images above are a little off from the final version. Here's the updated version:

Small cell size:
image


Large cell size (wider screenshot to show lineup):
image

gotdairyya (approved):
Code logic looks solid and tested out a few queries, also works really well! Thanks Jack!

PR #331, opened 2021-09-01
Make lineup re-render to adjust rowHeight based on cellSize

### LINKED ISSUE #323: Make lineup size dynamically with cellSize
Remnant from #320

JackWilb:
Closes #323

This PR adds a watcher for cellSize in the lineup component that redraws the whole lineup componenet. This comes with significant performance issues.

My goal is to explore lineup's documentation + implementation to see if it's possible to update the `rowHeight` parameter without a full re-render.

### DIFF
--- a/src/components/LineUp.vue
+++ b/src/components/LineUp.vue
@@ -13,6 +13,7 @@ export default defineComponent({
     const network = computed(() => store.state.network);
     const selectedNodes = computed(() => store.state.selectedNodes);
     const hoveredNodes = computed(() => store.state.hoveredNodes);
+    const cellSize = computed(() => store.state.cellSize);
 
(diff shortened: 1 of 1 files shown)


JackWilb:
After exploring the docs and communicating with the Caleydo team, I found lineup could adjust row height without tearing down and re-rendering. The relevant function is `dynamicHeight()` and to trigger the re-draw in lineup you have to call `lineup.update()`.

Small cell size:
image

Large cell size:
image

gotdairyya (approved):
this is great, thanks Jack. The only UI thing that sticks out is that the font size of lineup doesn't change and begins to look a little funny beyond the 15pt for cell size selection. But that's not really a bigger issue.

PR #337, opened 2021-09-14
Path table update

### LINKED ISSUE #334: Add path table to 1 hop query
On click of cell, generate plot table for 1 hop queries

Currently the path table is only shown for 2+ hops

gotdairyya:
With this update, you should be able to click the matrix (after a 1-hop query) and it will display a path table with the paths.

Closes #334

### DIFF
--- a/src/components/MultiMatrix.vue
+++ b/src/components/MultiMatrix.vue
@@ -3,6 +3,7 @@ import {
   Cell,
   Edge,
   Node,
+  ArangoPath,
 } from '@/types';
 import {
   scaleBand,
@@ -35,6 +36,7 @@ export default defineComponent({
 
   setup() {
     const showPathTable = computed(() => store.state.showPathTable);
+    const connectivityMatrixPaths = computed(() => store.state.connectivityMatrixPaths);
     const tooltip = ref(null);
     const visMargins = ref({
       left: 75, top: 110, right: 0, bottom: 0,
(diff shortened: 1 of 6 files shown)


src/components/MultiMatrix.vue
```diff
@@ -781,7 +791,9 @@ export default defineComponent({
           hideToolTip();
           unHoverEdge(matrixElement);
         })
-        .on('click', (event: MouseEvent, matrixElement: Cell) => store.commit.clickCell(matrixElement))
+        .on('click', (event: MouseEvent, matrixElement: Cell) => {
+          store.commit.clickCell(matrixElement);
+        })
```
JackWilb, line 796: This removes your on click logic you added below from the cells when they update again.

gotdairyya: Oh so I should add the logic in both parts right?

JackWilb: I believe so. I think not having it both places was causing it to be removed when the network updated for me.

src/components/PathTable.vue
```diff
@@ -131,6 +137,10 @@ export default defineComponent({
           });
         }
       });
+      // // Removes extra edge in even paths
+      // if (pathLength.value % 2 === 0) {
+      //   toReturn.pop();
+      // }
```
JackWilb, line 143: Looks like a remnant we can remove
```suggestion
```

JackWilb (changes requested):
Just one thing I saw. Everything else looks ready to go!

JackWilb (approved):

PR #343, opened 2021-11-17
Add additional query operators

### LINKED ISSUE #284: ConnectivityMatrix numerical queries
Add numerical querying for edges and nodes

>, <, !=, ==, <=, >=

JackWilb:
Closes #284

Adds `!=, <, <=, >, >=` query operators and allows them to work with numeric data types in arango. I explicityly cast `TO_NUMBER` when we're using a greater than or less than.

Please test and let me know what I broke. I did some minimal testing locally without any issue, though

### DIFF
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -130,29 +130,40 @@ export default defineComponent({
       });
     });
 
+    function isTextComparison(operator: string) {
+      return ['==', '=~'].includes(operator);
+    }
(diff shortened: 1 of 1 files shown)


JackWilb:
> It worked for me too. The only issue I could see if someone tries to use the `=~` operator on a number -- It's unclear what happens then

Yeah, that's true. That's the case with arango in general, though, since `=~` regex match. We'll just have to document it when we get around to that 😄

PR #356, opened 2022-02-02
Allow label update

JackWilb:
No. We hadn't opened one for it

Add the ability to toggle the label variable between any of the node attributes. If no label is selected, the `_key` is shown

Without a variable selected:
image

With a variable selected:
image

TODO:
- [x] Make an issue to refactor controls. I plan to make it more like the multilink one (#357)

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -8,7 +8,7 @@ import AboutDialog from '@/components/AboutDialog.vue';
 import LoginMenu from '@/components/LoginMenu.vue';
 import ConnectivityQuery from '@/components/ConnectivityQuery.vue';
 import {
-  computed, defineComponent, ref, watchEffect,
+  computed, defineComponent, ref, watch, watchEffect,
 } from '@vue/composition-api';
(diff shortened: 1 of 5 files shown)


gotdairyya (dismissed):
Looks good (and happy for this functionality). It does not play well with the aggregated node types though:
- the label is funky after nodes are aggregated
- the label doesn't show immediately if a label was chosen but then the nodes are aggregated

I will go ahead and approve, let me know if you want me to create issues for this.

JackWilb:
> Looks good (and happy for this functionality). It does not play well with the aggregated node types though:

I think I've addressed some of that. Would you mind double checking it? I updated it so that when you aggregate, the variable you aggregate by is chosen as the label. I have some ideas for how we might allow different values for the supernodes and the children.

> I will go ahead and approve, let me know if you want me to create issues for this.

I'll create an issue for labeling with different values for children and supernodes once you check my changes.

gotdairyya (approved):
Works great!

PR #359, opened 2022-02-02
Remove internal columns from lineup, except _key

### LINKED ISSUE #358: Hide internal variables from lineup

JackWilb:
Depends on #356 

Closes #358

This PR hides internal variables from lineup. I added back `_key`, since it's the default label variable, and I didn't want it to be confusing

Here's what remains on the les mis data set:
image

No

### DIFF
--- a/src/components/LineUp.vue
+++ b/src/components/LineUp.vue
@@ -5,6 +5,7 @@ import {
 } from '@vue/composition-api';
 import LineUp, { DataBuilder } from 'lineupjs';
 import { select } from 'd3-selection';
+import { isInternalField } from '@/lib/typeUtils';
 
 export default defineComponent({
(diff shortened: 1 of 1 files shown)


JackWilb:
The build will fail until this PR is rebased onto main after #356

gotdairyya (approved):
When the matrix is aggregated lineup shows a clunky string from objects, it might be better to replace that with a `-`:
image

Otherwise it looks great and functions as expected!

JackWilb:
> When the matrix is aggregated lineup shows a clunky string from objects, it might be better to replace that with a `-`

You must have had an older version of the branch checked out. `children` is an internal variable, now, and is not shown in lineup. Here's some visual proof 😄 
image

PR #364, opened 2022-02-03
Re-implement node searching

### LINKED ISSUE #363: Implement search with search box in the side panel
This should look like multilink's for consistency across the apps

JackWilb:
Closes #363 

I re-implemented node searching, using similar UI and functions to those that exist in multilink. This new search bar acts as if you're clicking. That is, if the node is unselected it will be selected, and if it's selected it will be unselected. This might not be ideal, but to change it would require a rewrite of quite a bit of code. Let me know how the UX is. The search also works within aggregated nodes.

I didn't test with any intermediary nodes, so that might be broken.

I also adjusted some margins that weren't needed after this change

New search bar:
image

TODOs:
- [x] Test with intermediary nodes table

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -10,7 +10,6 @@ import ConnectivityQuery from '@/components/ConnectivityQuery.vue';
 import {
   computed, defineComponent, ref, watch, watchEffect,
 } from '@vue/composition-api';
-import { internalFieldNames, Node } from '@/types';
 
 export default defineComponent({
   components: {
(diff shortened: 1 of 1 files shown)


gotdairyya (approved):
Works with the intermediary nodes. One behavior that's unexpected though is that if you search for another node then both rows will be selected -- I'm not sure if that's the behavior you are envisioning. 

Search for Child1:
image

Search for Child5 right after:
image

Both rows end up being selected.

JackWilb:
> Works with the intermediary nodes. One behavior that's unexpected though is that if you search for another node then both rows will be selected -- I'm not sure if that's the behavior you are envisioning.
> Both rows end up being selected.

Yeah, I thought it might be a bit weird. I'll make an issue

PR #366, opened 2022-02-04
Selected nodes in store

### LINKED ISSUE #365: Fix node searching by putting selected nodes into the store
> Works with the intermediary nodes. One behavior that's unexpected though is that if you search for another node then both rows will be selected -- I'm not sure if that's the behavior you are envisioning.
> Both rows end up being selected.

_Originally posted by @jackwilb in https://github.com/multinet-app/multimatrix/issues/364#issuecomment-1029538402_

JackWilb:
Closes #365

This is all under the hood changes and shouldn't have an effect on the UI. Apparently the selected nodes were already tracked in the store so I changed the variable tracking them to a set, since they should be unique.

Additionally, I added some helper functions that are used in multilink to help speed up some of my future work. This will mean I can copy some code over verbatim to clear the selection, etc.

No behavior changes expected.

No

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -186,7 +186,7 @@ export default defineComponent({
           .map((node) => node._id);
 
         if (nodeIDsToSelect.length > 0) {
-          nodeIDsToSelect.forEach((id) => store.commit.clickElement(id));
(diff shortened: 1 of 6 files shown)


gotdairyya (approved):
Looks great!

PR #368, opened 2022-02-04
Only allow aggregating by category if column types are available

### LINKED ISSUE #360: Allow a different label for children and super nodes
I'll create an issue for labeling with different values for children and supernodes once you check my changes.

_Originally posted by @JackWilb in https://github.com/multinet-app/multimatrix/issues/356#issuecomment-1028477195_

### LINKED ISSUE #376: Column types from one table are overwritten by another table with the same variable name
For example:

Nodes:
```
...
    x: "category"
...
```

Links:
```
...
    x: "label"
...
```

In this case, it's not clear who will win, and we probably want both definitions regardless

JackWilb:
Closes #360 (thanks @gotdairyya)
Closes #376

This PR restricts the columns we can aggregate by if we have type information from the API. If the type data is null, it will show all columns as before.

N/A

TODOs:
- [x] Test how this works with marclab data
- [x] See if empty metadata comes back as null or `{}` (every table must be uploaded with metadata)

### DIFF
--- a/src/components/Alert.vue
+++ b/src/components/Alert.vue
@@ -57,7 +57,7 @@ export default defineComponent({
       networkOptions,
       workspace,
       workspaceOptions,
-      edgeTableName: computed(() => store.state.edgeTableName),
+      edgeTableName: computed(() => store.getters.edgeTableName),
     };
   },
 });
--- a/src/components/Alert.vue
+++ b/src/components/Alert.vue
@@ -57,6 +57,7 @@ export default defineComponent({
       networkOptions,
       workspace,
       workspaceOptions,
+      edgeTableName: computed(() => store.state.edgeTableName),
     };
   },
 });
@@ -123,7 +124,7 @@ export default defineComponent({
     </v-alert>
 
     <filter-overlay
-      v-if="loadError.message === 'The network you are loading is too large'"
+      v-if="loadError.message === 'The network you are loading is too large' && edgeTableName !== null"
     />
   </div>
 </template>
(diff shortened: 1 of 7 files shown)


gotdairyya (approved):
Works except I've noticed two bugs.

I've recorded one as issue #369 

The other is that the connectivity querying is no longer working for me on my local builds -- @JackWilb can you check if this is a me problem or whether it is actually broken?

JackWilb:
This seems to break with marclab data, so let's figure out how to fix that before we merge @gotdairyya

gotdairyya:
> This seems to break with marclab data, so let's figure out how to fix that before we merge @gotdairyya

The query works, but the aggregation then is broken:
image

gotdairyya:
So I tested locally with RPC1 data and it works as expected:
image

JackWilb:
Changes in https://github.com/multinet-app/marclab/pull/6 should fix the issue with the marclab. That should unblock this from merging @gotdairyya

gotdairyya (approved):
Looks good!

JackWilb:
The problem here is with the columnTypes. This is going to require a whole re-write of the code that handles the types, so this will be on hold until that's done. Tracking that in #376

gotdairyya:
Kind of an interesting behavior. When aggregating a queried set, lineup does not show the extra columns that appear when a parent node is expanded

image

versus
image

gotdairyya (dismissed):
Just submitted a commit that syncs the label and aggregated variable and also updates the matrix so that two different labels can be shown. Not sure what to do about the comment I put above, what do you think?

JackWilb:
> Kind of an interesting behavior. When aggregating a queried set, lineup does not show the extra columns that appear when a parent node is expanded

I think this is fine behavior. The supernodes don't have those columns so it would just be a column of undefined for each attribute. We could perhaps show some kind of aggregated metrics for each (averages, min, maxes, etc.)

gotdairyya (approved):

PR #370, opened 2022-02-10
Make querying more robust

### LINKED ISSUE #282: ConnectivityMatrix and/or edge functionality
Moving part of issue #278 here because adding an AND/OR functionality on the edges will take some more work

### LINKED ISSUE #335: Add advanced querying to nodes
Allow option to expand node querying in connectivity matrix to include OR/NOT*

Examples to illustrate each:

- I want to see flights from (Boston OR Chicago) to (NYC)
- I want to see flights from (San Jose NOT San Francisco) to (SLC)

If we add an AND to the query options it will have to be about another attribute, for example:

- I want to see connections between (Characters in chapter 1 AND women) to (Characters in chapter 3)

gotdairyya:
Closes #335 
Closes #282 

This PR will make it so that a user can conduct advanced querying of the connectivity network.
At the moment we can only query for one attribute per node or edge, and we'd like to query for more than one attribute.

TODOs:
- [x] Create UI for expanding + collapsing querying
- [x] Update AQL queries to work with multi-attribute querying

### DIFF
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -43,7 +43,7 @@
         <!-- All of the query options -->
         <v-list dense>
           <v-list-item
-            v-for="(val, j) in inputs.value[0]"
+            v-for="(val, j) in inputs.value"
             :key="`val-${i}-${j}`"
             class="pa-0"
           >
@@ -143,6 +143,7 @@
             </v-list-item>
           </v-list-item>
         </v-list>
+
         <!-- Let user query for another edge -->
         <v-card-actions v-if="i === 1">
           <v-btn
@@ -171,7 +172,7 @@
           <div v-show="showSecondEdge && i === 1">
             <v-list dense>
               <v-list-item
-                v-for="(val, k) in inputs.value[i]"
+                v-for="(val, k) in edgeMutexs.value"
                 :key="`val-${i}-2-${k}`"
                 class="pa-0"
               >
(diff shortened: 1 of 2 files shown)


JackWilb:
I've updated the TODOs above, but we should chat about how the `NOT` logic should work, and if we should allow for paths that go through the start node again (e.g. the path goes out and back and then to a target)

gotdairyya:
The test query for the marclab crashes the browser for the RC1 dataset, but works for RPC1 (which is smaller):

`(CBb*) -[BC Conventional Synapse] AND NOT [Ribbon Synapse]-> (*)`

gotdairyya:
@JackWilb the UI is 🗑️🔥 right now and the new data structure might be too. But it's a start, let's chat sometime this week to see what improvements can be made

gotdairyya:
For context, Jack and I met to discuss how there are two different types of edge queries that happen and how the first iteration of the UI could support both independently but not together. 

The two types of edge queries are:
1. Querying along 1 edge but using that edge's attributes as filters
2. Querying along multiple edges between nodes

At the moment, given the use case with the marclab we are only allowing the user to search along the first hop and only two different edges. We figured this is a good compromise given that we don't have other user cases and it will increase the complexity of the AQL query.

Here is our sketch for the querying:
image=

@JackWilb please add to this if you think I am missing anything

JackWilb (approved):

PR #372, opened 2022-02-10
Edge slicing

gotdairyya:
No.

This is incorporating the functionality of MultiDynamic into MultiMatrix. It will be more generalized so that it can also account for categorical edge types.

Edge slicing lets you pick numerical and date-formated edge variables and create slices.
image

Once the slices are created, you can navigate the network along the slice
image

TODOs:
- [x] Check why variables don't load for networks that need subsetting
- [x] Add metadata to EdgeSlices.vue (toopltip + visualization)
- [x] Fix type issues
- [x] Clear timeline view for the hop query

### DIFF
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -228,6 +228,8 @@ export default defineComponent({
             store.dispatch.updateNetwork({ network: newNetwork });
             loading.value = false;
             store.commit.setDirectionalEdges(true);
+            // remove sliced network
+            store.commit.setSlicedNetwork([]);
           }
         });
       }
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -1,87 +1,112 @@
 <template>
   <div class="pa-0">
+    <v-subheader class="grey darken-3 py-0 pr-0 white--text">
+      Connectivity Query
+
+      <v-spacer />
+
+      <v-btn
+        :min-width="40"
+        :height="48"
+        depressed
+        tile
+        class="grey darken-3 pa-0"
+        @click="showMenu = !showMenu"
+      >
+        <v-icon color="white">
+          {{ showMenu ? 'mdi-chevron-up' : 'mdi-chevron-down' }}
+        </v-icon>
+      </v-btn>
+    </v-subheader>
+
     <v-card
-      color="grey lighten-3"
+      v-if="showMenu"
       flat
       tile
     >
-      <v-card-text class="pt-2 pb-1">
-        <v-select
-          v-model="selectedHops"
-          label="Hops"
-          :items="hopsSelection"
-        />
-      </v-card-text>
-    </v-card>
-    <v-list dense>
-      <v-list-item
-        v-for="(_, i) in displayedHops"
-        :key="i"
-        class="pa-0"
+      <v-card
+        color="grey lighten-3"
+        flat
+        tile
       >
-        <v-list-item-avatar class="mr-0">
-          <v-icon size="18">
-            {{ i % 2 ? 'mdi-swap-vertical' : `mdi-numeric-${(i+2)/2}-circle` }}
-          </v-icon>
-        </v-list-item-avatar>
-        <v-list-item-content class="pa-0 pr-1">
-          <v-row no-gutters>
-            <v-col
-              cols="12"
-              sm="5"
-              class="pa-1"
-            >
-              <v-autocomplete
-                v-model="selectedVariables[i]"
-                :items="i % 2 ? edgeVariableItems : nodeVariableItems"
-                dense
-              />
-            </v-col>
-            <v-col
-              cols="12"
-              sm="3"
-              class="pa-1"
-            >
-              <v-autocomplete
-                v-model="selectedQueryOptions[i]"
-                :items="queryOptionItems"
-                dense
-              />
-            </v-col>
-            <v-col
-              cols="12"
-              sm="4"
-              class="pa-1"
-            >
-              <v-autocomplete
-                v-if="selectedQueryOptions[i] === '=='"
-                v-model="selectedVariableValue[i]"
-                :items="variableValueItems[i]"
-                dense
-              />
-              <v-text-field
-                v-else
-                v-model="selectedVariableValue[i]"
-                dense
-              />
-            </v-col>
-          </v-row>
-        </v-list-item-content>
-      </v-list-item>
-
-      <v-list-item>
-        <v-btn
-          block
-          class="ml-0 mt-4"
-          color="primary"
-          depressed
-          :loading="loading"
-          @click="submitQuery"
+        <v-card-text class="pt-2 pb-1">
+          <v-select
+            v-model="selectedHops"
+            label="Hops"
+            :items="hopsSelection"
+          />
+        </v-card-text>
+      </v-card>
+      <v-list dense>
+        <v-list-item
+          v-for="(_, i) in displayedHops"
+          :key="i"
+          class="pa-0"
         >
-          Submit Query
-        </v-btn>
-      </v-list-item>
-    </v-list>
+          <v-list-item-avatar class="mr-0">
+            <v-icon size="18">
+              {{ i % 2 ? 'mdi-swap-vertical' : `mdi-numeric-${(i+2)/2}-circle` }}
+            </v-icon>
+          </v-list-item-avatar>
+          <v-list-item-content class="pa-0 pr-1">
+            <v-row no-gutters>
+              <v-col
+                cols="12"
+                sm="5"
+                class="pa-1"
+              >
+                <v-autocomplete
+                  v-model="selectedVariables[i]"
+                  :items="i % 2 ? edgeVariableItems : nodeVariableItems"
+                  dense
+                />
+              </v-col>
+              <v-col
+                cols="12"
+                sm="3"
+                class="pa-1"
+              >
+                <v-autocomplete
+                  v-model="selectedQueryOptions[i]"
+                  :items="queryOptionItems"
+                  dense
+                />
+              </v-col>
+              <v-col
+                cols="12"
+                sm="4"
+                class="pa-1"
+              >
+                <v-autocomplete
+                  v-if="selectedQueryOptions[i] === '=='"
+                  v-model="selectedVariableValue[i]"
+                  :items="variableValueItems[i]"
+                  dense
+                />
+                <v-text-field
+                  v-else
+                  v-model="selectedVariableValue[i]"
+                  dense
+                />
+              </v-col>
+            </v-row>
+          </v-list-item-content>
+        </v-list-item>
+
+        <v-list-item>
+          <v-btn
+            block
+            class="ml-0 mt-4"
+            color="primary"
+            depressed
+            :loading="loading"
+            @click="submitQuery"
+          >
+            Submit Query
+          </v-btn>
+        </v-list-item>
+      </v-list>
+    </v-card>
   </div>
 </template>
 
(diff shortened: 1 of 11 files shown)


src/components/EdgeSlicing.vue
```diff
@@ -0,0 +1,398 @@
...
+        <v-btn
+          :min-width="40"
+          :height="48"
+          depressed
+          tile
+          :class="showOptions? `grey darken-2 pa-0` : `grey darken-3 pa-0`"
+          @click="showOptions = !showOptions"
+        >
+          <v-icon color="white">
+            mdi-cog
```
JackWilb, line 233: Let's use mdi-chevron-up/mdi-chevron-down here. Similar to how it is for visualization options in the main controls component

src/store/index.ts
```diff
@@ -9,14 +9,15 @@ import { initProvenance, Provenance } from '@visdesignlab/trrack';
 import api from '@/api';
 import oauthClient from '@/oauth';
 import {
+  ColumnTypes,
   NetworkSpec, UserSpec,
```
JackWilb, line 13: Condense into one line

JackWilb (approved):
Looks good so far, the UI needs some work, as you said. I think it's good to go, but maybe we should hold off until you have a PR for the UI changes.

Also, is there anything different in the edge slicing, and edge slices component that I haven't review in the multidyno?

src/components/EdgeSlicing.vue
```diff
@@ -0,0 +1,398 @@
...
+
+      list.forEach((variable) => {
+        if (columnTypes.value !== null && columnTypes.value[variable] !== 'label') {
+          cleanedVariables.add(variable);
+        }
+      });
+      return cleanedVariables;
+    }
+
+    const cleanedEdgeVariables = computed(() => {
```
JackWilb, line 53: Can these be moved to the store, we have similar logic for the node variable lists there. It would make sense to put similar code together, and maybe we can even reduce some of the logic needed

JackWilb:
> @JackWilb can you take a look at the to-do I listed? I think the reason it doesn't work is because of when the edge variable list is created. It might need to be generated `onMounted()`

It seems that the list generation relies on `networkOnLoad`, which wouldn't exist if the network doesn't get fetched because it's too big. You might need a refactor of that code if you want to see the list before you have a network downloaded. Also, it would then have to fetch the network split by time slices

src/components/EdgeSlicing.vue
gotdairyya: done

src/store/index.ts
gotdairyya: done

gotdairyya:
> Looks good so far, the UI needs some work, as you said. I think it's good to go, but maybe we should hold off until you have a PR for the UI changes.
> 
> Also, is there anything different in the edge slicing, and edge slices component that I haven't review in the multidyno?

nope! I just refactored some of the code so that it works with the MultiMatrix setup, but nothing new yet

gotdairyya:
Made some bigger changes:
- slices show the number of edges within each section
- now slices can be made along categorical variables
- cleaned up the UI stuff that was bothering me
- slices are cleared when a connectivity query is made

Left to do: I need some help with the type errors (sorry, I just can't figure out what's going on this time around)

src/components/ConnectivityQuery.vue
```diff
@@ -228,6 +228,8 @@ export default defineComponent({
             store.dispatch.updateNetwork({ network: newNetwork });
             loading.value = false;
             store.commit.setDirectionalEdges(true);
+            // remove sliced network
+            store.commit.setSlicedNetwork([]);
```
JackWilb, line 232: Maybe this should just be a part of `updateNetwork`? It seems that every time it updates, we'd want to unset this

JackWilb (changes requested):
I'm planning to give you a couple rounds of review. The first here will be higher level issues, and then I'll comment on specific code that might be updated in the next round. 

Here are a few other things I noticed:

- I can't see a tooltip
- It was unintuitive to me that some of the slices were orange and some were not when I first made a network. The tooltip will help with this, but is there some other way we can show what the orange represents?
- There isn't enough padding on the "Edge Slices" at the top of the vis
- It looks like the SVG you added could be smaller from some playing around. I think 30px high would make it fit in a bit more. Unless I missed something that goes below each of the bars.
- I found it unintuitive that there were separate variables for start and end. Looking at your screenshot, it seems a bit more clear, but why would we want it across 2 variables? Is that something we could hide unless the user asks for it, knowing that `end variable` is a separate column in their data?
- There should be a button somewhere that will clean all the edge slicing away. It seems that the best way to do that currently is to remove the variables from the panel and re-run the slicing, but that's too much work for a user. If it's one button to clear, that will be much nicer.
- We should open an issue to make sure that the sorting we implement is able to take information from slices instead of the whole network
- Is there a nice way we can show what ranges of values fall into each time slice?

gotdairyya:
Date + Time View:
image

Numeric + Time View:
image

Numeric View:
image

Categorical View:
image

src/components/EdgeSlicing.vue
```diff
@@ -0,0 +1,446 @@
...
+    });
+
+    watch([dateFormatted], () => {
+      if (isDate.value) {
+        varRange.value[0] = new Date(dateFormatted.value[0]);
+        varRange.value[1] = new Date(dateFormatted.value[1]);
+      }
+    });
+
+    function sliceNetwork() {
```
JackWilb, line 96: This fails silently if the range specified by user is outside of the values of the data. For example, I have data from 0 to 31 and if I specify -100 to 31, it fails without alerting the user. We could have some simple validation and show errors to the user when it happens

JackWilb (changes requested):
Looking good so far! I have a smaller round of changes here that will just help with some papercuts

src/components/FilterOverlay.vue
```diff
@@ -161,6 +161,7 @@ export default defineComponent({
           if (aqlNetwork.nodes.length !== 0) {
           // Update state with new network
             store.dispatch.updateNetwork({ network: aqlNetwork });
+            store.commit.setNetworkOnLoad(aqlNetwork);
```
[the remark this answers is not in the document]
gotdairyya: Done

src/components/EdgeSlicing.vue
```diff
@@ -0,0 +1,446 @@
...
+                @click="sliceNetwork"
+              >
+                Generate Slices
+              </v-btn>
+            </v-col>
+            <v-col
+              cols="12"
+              sm="6"
+            >
+              <v-btn
```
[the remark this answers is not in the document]
gotdairyya: Done

src/components/EdgeSlices.vue
```diff
@@ -0,0 +1,271 @@
...
+        tooltipMessage.value = `Category: ${currentSlice.value.categories[key]}
+      Total Edges: ${currentSlice.value.sumEdges[key]}`;
+        toggleTooltip.value = true;
+      }
+    }
+
+    function hideTooltip() {
+      tooltipMessage.value = '';
+      toggleTooltip.value = false;
+    }
```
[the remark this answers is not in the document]
gotdairyya: Issue #379

src/components/EdgeSlices.vue
```diff
@@ -0,0 +1,271 @@
...
+
+export default defineComponent({
+  name: 'EdgeSlices',
+
+  setup() {
+    const slicedNetwork = computed(() => store.state.slicedNetwork);
+    const isDate = computed(() => store.state.isDate);
+    const isNumeric = computed(() => (slicedNetwork.value[0].category === ''));
+    const currentInstance = getCurrentInstance();
+    const svgWidth = computed(() => (currentInstance !== null ? currentInstance.proxy.$vuetify.breakpoint.width - store.state.controlsWidth : 0));
```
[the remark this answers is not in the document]
gotdairyya: Issue #380

src/components/EdgeSlices.vue
```diff
@@ -0,0 +1,271 @@
...
+        slices.sumEdges[i] = slice.network.edges.length;
+        if (isNumeric.value) {
+          slices.timeRanges[i] = slice.time;
+        } else {
+          slices.categories[i] = slice.category;
+        }
+      });
+      return slices;
+    });
+    const timeExtent = computed(() => extent(Object.values(currentSlice.value.timeRanges).flat()));
```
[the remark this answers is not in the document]
gotdairyya: Done

src/components/EdgeSlices.vue
```diff
@@ -0,0 +1,271 @@
...
+    // Select the first slice on initial load
+    onMounted(() => {
+      select('#edgeSlice_0').classed('selected', true);
+    });
+
+    // Select the first slice on when slices are changed load
+    watch([slicedNetwork], () => {
+      selectAll('.edgeSliceRectClass').classed('selected', false);
+      select('#edgeSlice_0').classed('selected', true);
+    });
```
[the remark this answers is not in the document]
gotdairyya: Done

src/components/EdgeSlices.vue
```diff
@@ -39,6 +38,14 @@ export default defineComponent({
 
     const timeRangesLength = computed(() => currentSlice.value.slices);
 
+    // Update sliced view and network
+    const selectedArray: ComputedRef<boolean[]> = computed(() => Array.from(Array(timeRangesLength.value), (_, x: number) => (x === 0)));
+    function isSelected(key: number) {
+      selectedArray.value.fill(false);
+      selectedArray.value[key] = true;
```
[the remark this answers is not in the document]
waxlamp: FWIW (might not apply here) you can use a "computed setter" (V2: https://v2.vuejs.org/v2/guide/computed.html#Computed-Setter) or a "writable computed" (V3: https://vuejs.org/guide/essentials/computed.html#writable-computed) to give semantics to "changing a value on a computed ref".

src/components/EdgeSlicing.vue
gotdairyya: resolved for both number inputs _and_ date inputs

src/components/EdgeSlices.vue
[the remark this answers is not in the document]
gotdairyya: Ok, this has also been fixed!

gotdairyya:
As an overview, here is the expected views by data type.

**CATEGORICAL**
image

**TIME BUT NOT DATE**
image

**TIME AND DATE**
image

src/components/EdgeSlicing.vue
```diff
@@ -490,7 +491,7 @@ export default defineComponent({
                 color="primary"
                 block
                 depressed
-                :disabled="!isValidRange && !isDate"
+                :disabled="!isValidRange && !isDate && isNumeric"
```
JackWilb, line 494: Is this right? Shouldn't it be just `!isValidRange`? Why would we disable it if it's numeric?

src/components/EdgeSlicing.vue
```diff
@@ -15,14 +19,15 @@ export default defineComponent({
     const showMenu = ref(false);
     const sliceRules = (value: string) => !Number.isNaN(parseFloat(value)) || 'Please type a number';
     const calMenu = ref([false, false]);
-    const dateFormatted: Ref<string[]> = ref([]);
+    // const dateFormatted: Ref<string[]> = ref([]);
```
JackWilb, line 22: Remnant

JackWilb (approved):
Noticed a couple of small things, but in general, it's looking good to go!

src/components/EdgeSlicing.vue
gotdairyya: This is so that the `isValidRange` only does the check for `isNumeric` selections. Otherwise, you can't slice along categories

PR #383, opened 2022-03-09
Refactor int node vis

gotdairyya:
No

This code refactors the intermediary node table. This makes the code more legible and also clears up some of the visual glitches that came from updating. It also fixed a minor issue in displaying edges in the path table component.

No.

### DIFF
--- a/src/components/IntermediaryNodes.vue
+++ b/src/components/IntermediaryNodes.vue
@@ -1,11 +1,11 @@
 <script lang="ts">
 import {
-  computed, defineComponent, onMounted, watch, watchEffect,
+  computed, defineComponent, watchEffect,
 } from '@vue/composition-api';
 import {
   scaleLinear,
 } from 'd3-scale';
-import { select, selectAll } from 'd3-selection';
+import { selectAll } from 'd3-selection';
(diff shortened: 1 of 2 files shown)


src/components/IntermediaryNodes.vue
```diff
@@ -233,5 +235,6 @@ svg >>> .connectivityCell:hover {
   cursor: pointer;
   stroke-width: 1px;
   stroke: black;
+  z-index: 100;
```
JackWilb, line 238: This doesn't work, they still get covered over. I think it looks better without this fix or we'll need to spend more time figuring out how to do this so the border works on hover properly.

Given, this might be a firefox issue only

JackWilb (changes requested):
I noticed that the path table is still not showing the right number of nodes. There should be 4 in this table:

image

src/components/IntermediaryNodes.vue
gotdairyya: I am removing it

gotdairyya:
An important note. Jack and I decided to limit the max hops to 3. There aren't any use cases at the moment that require more than three hops.

PR #389, opened 2022-05-09
Calc degrees

### LINKED ISSUE #386: Filter out empty nodes
Filter empty nodes throughout the edge slice view.

gotdairyya:
Closes #386 

Let's the user filter nodes based on the number of out degrees.

I'm sure the code will need some sprucing up. I'm not sure I computed the filtering in the best way.
TODOs:
- [x] Review code for filtering
- [x] Check how this works with aggregated networks and queried networks

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -445,24 +445,20 @@ export default defineComponent({
               color="blue darken-1"
             >
               <template v-slot:prepend>
-                <v-text-field
-                  v-model="degreeRange[0]"
-                  class="pa-0 ma-0"
-                  hide-details
-                  single-line
-                  type="number"
-                  style="width: 30px"
-                />
+                <p
+                  class="pa-0 ma-0 text-center"
+                  style="min-width: 25px; color: rgba(255, 255, 255, 0.7);"
+                >
+                  {{ degreeRange[0] }}
+                </p>
               </template>
               <template v-slot:append>
-                <v-text-field
-                  v-model="degreeRange[1]"
-                  class="pa-0 ma-0"
-                  hide-details
-                  single-line
-                  type="number"
-                  style="width: 30px"
-                />
+                <p
+                  class="pa-0 ma-0 text-center"
+                  style="min-width: 25px; color: rgba(255, 255, 255, 0.7);"
+                >
+                  {{ degreeRange[1] }}
+                </p>
               </template>
             </v-range-slider>
           </v-list-item>
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -3,6 +3,7 @@ import { select } from 'd3-selection';
 import { format } from 'd3-format';
 import { legendColor } from 'd3-svg-legend';
 import { ScaleLinear } from 'd3-scale';
+import { max } from 'd3-array';
 import store from '@/store';
 import AboutDialog from '@/components/AboutDialog.vue';
 import LoginMenu from '@/components/LoginMenu.vue';
@@ -81,33 +82,38 @@ export default defineComponent({
     const maxConnections = computed(() => store.state.maxConnections);
 
     // Calculate node degrees
-    const degreeMax = computed(() => (networkOnLoad.value !== null ? networkOnLoad.value.nodes.length : 0));
-    const degreeRange = computed({
-      get() {
-        return store.state.degreeRange;
-      },
-      set(range: number[]) {
-        store.commit.setDegreeRange(range);
-      },
+    // Create dict of row nodes and max degrees
+    const nodeDegreeDict: {[key: string]: {[key: string]: number }} = {};
+
+    const degreeMax = computed(() => {
+      let maxCounter = 0;
+      if (networkOnLoad.value != null) {
+        networkOnLoad.value.nodes.forEach((node) => {
+        // eslint-disable-next-line no-return-assign, no-sequences, no-param-reassign
+          nodeDegreeDict[node._id] = node.neighbors.reduce((cnt: string, cur: string) => (cnt[cur] = cnt[cur] + 1 || 1, cnt), {});
+          const currentMax = max(Object.values(nodeDegreeDict[node._id]));
+          if (currentMax !== undefined && currentMax > maxCounter) {
+            maxCounter = currentMax;
+          }
+        });
+      }
+      return maxCounter;
     });
+
+    const degreeRange = ref([0, degreeMax.value]);
 
     function removeByDegree() {
       // Restore network if min and max are restored
-      if (networkOnLoad.value !== null && degreeRange.value[0] === 0 && degreeRange.value[1] === networkOnLoad.value.nodes.length) {
+      if (networkOnLoad.value !== null && degreeRange.value[0] === 0 && degreeRange.value[1] === degreeMax.value) {
         store.dispatch.updateNetwork({ network: networkOnLoad.value });
-      } else {
+      } else if (networkOnLoad.value !== null) {
         // eslint-disable-next-line no-undef
         const newNetwork = structuredClone(networkOnLoad.value);
         const nodeSet = new Set([]);
 
-        // Create dict of row nodes and max degrees
-        const nodeDegreeDict: {[key: string]: number } = {};
-        // eslint-disable-next-line no-return-assign
-        newNetwork.nodes.forEach((node: Node) => nodeDegreeDict[node._id] = node.degreeCount);
-
         // Remove edges that don't match degree criteria
         newNetwork.edges = newNetwork.edges.filter((edge: Edge) => {
-          if (nodeDegreeDict[edge._from] >= degreeRange.value[0] && nodeDegreeDict[edge._from] <= degreeRange.value[1]) {
+          if (nodeDegreeDict[edge._from][edge._to] >= degreeRange.value[0] && nodeDegreeDict[edge._from][edge._to] <= degreeRange.value[1]) {
             nodeSet.add(edge._from);
             nodeSet.add(edge._to);
             return true;
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -195,7 +195,9 @@ export default defineComponent({
     function aggregateNetwork(varName: string) {
       if (filtered.value) {
         store.commit.setFilteredNetwork(false);
-        if (networkOnLoad.value !== null) { store.dispatch.updateNetwork({ network: networkOnLoad.value }); }
+        if (networkOnLoad.value !== null) {
+          store.dispatch.updateNetwork({ network: networkOnLoad.value });
+        }
         degreeRange.value = [0, maxDegree.value];
         store.commit.setDegreeNetwork(degreeRange.value);
       }
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -65,6 +65,7 @@ export default defineComponent({
       },
     });
     const aggregated = computed(() => store.state.aggregated);
+    const filtered = computed(() => store.state.filteredNetwork);
     const cellColorScale = computed(() => store.getters.cellColorScale);
     const parentColorScale = computed(() => store.getters.parentColorScale);
     const nodeVariableItems = computed(() => store.getters.nodeVariableItems);
@@ -220,6 +221,7 @@ export default defineComponent({
       showGridLines,
       cellSize,
       aggregated,
+      filtered,
       cellColorScale,
       parentColorScale,
       nodeVariableItems,
@@ -466,11 +468,11 @@ export default defineComponent({
         <div class="pa-4">
           <!-- Aggregated Matrix Legend -->
           <v-list-item
-            v-if="aggregated"
+            v-if="aggregated || filtered"
             class="pb-0 px-0"
             style="display: flex; max-height: 50px"
           >
-            Aggregate Legend
+            {{ aggregated ? 'Aggregate Legend' : 'Filtered Legend' }}
             <svg
               id="parent-matrix-legend"
               height="50"
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -100,6 +100,7 @@ export default defineComponent({
 
     // Non-template objects
     const network = computed(() => store.state.network);
+    const networkOnLoad = computed(() => store.state.networkOnLoad);
 
     const searchTerm = ref('');
     const searchErrors: Ref<string[]> = ref([]);
@@ -192,6 +193,8 @@ export default defineComponent({
     }
 
     function aggregateNetwork(varName: string) {
+      store.commit.setFilteredNetwork(false);
+      if (networkOnLoad.value !== null) { store.dispatch.updateNetwork({ network: networkOnLoad.value }); }
       store.dispatch.aggregateNetwork(varName);
     }
 
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -11,6 +11,7 @@ import EdgeSlicing from '@/components/EdgeSlicing.vue';
 import {
   computed, defineComponent, Ref, ref, watch, watchEffect,
 } from '@vue/composition-api';
+import { Edge } from '@/types';
 
 export default defineComponent({
   components: {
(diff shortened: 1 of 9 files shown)


gotdairyya:
Also, @JackWilb I set the min and max attributes for the range-slider, but for some reason, they aren't limits (ex. if the min in 0, I can still input a value less than 0)

JackWilb (changes requested):
Looking good. I pushed a change for the slider UI (which I think will address your most recent comment) and I left some comments for how you might improve the filtering. I also noticed some general stuff that it would be best to fix up. Let me know if you'd like me to help with any of it

src/store/index.ts
```diff
@@ -456,13 +461,14 @@ const {
         edges: edges.results as Edge[],
       };
       commit.setAttributeValues(networkElements);
+      commit.setNetworkOnLoad(networkElements);
+      commit.setDegreeRange([0, networkElements.nodes.length]);
```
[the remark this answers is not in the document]
gotdairyya: I would normally agree but in my example network there is a node with no interactions for this specific dataset so by that metric (max degree) it would never show up

src/lib/utils.ts
```diff
@@ -26,6 +26,9 @@ export function defineNeighbors(nodes: any[], edges: Edge[]) {
     }
   });
 
+  // eslint-disable-next-line no-return-assign, no-param-reassign
+  nodes.forEach((d) => d.degreeCount = new Set(d.neighbors).size);
```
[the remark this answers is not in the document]
gotdairyya: > Is this how the degree count should work? If there are 100 edges between 2 people, they would have degree 1

Did I misinterpret what `d.neighbors` represents?


> Another issue is that we're tracking the same data in 2 different variables. It seems that degree count is just the size of the neighbors. Why don't we make neighbors a set or an object (to count edges, e.g. {"Jack": 10, "Derya": 20}), and then just use the size or length of keys instead of making a separate degreeCount variable?

Why would that be better? If it is a node variable then it will also show up in the UpSet table which could be useful.

src/components/Controls.vue
```diff
@@ -78,6 +79,62 @@ export default defineComponent({
     });
     const maxConnections = computed(() => store.state.maxConnections);
 
+    // Calculate node degrees
+    const degreeMax = computed(() => (networkOnLoad.value !== null ? networkOnLoad.value.nodes.length : 0));
```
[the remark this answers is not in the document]
gotdairyya: I will check

src/components/Controls.vue
```diff
@@ -78,6 +79,62 @@ export default defineComponent({
...
+        const nodeDegreeDict: {[key: string]: number } = {};
+        // eslint-disable-next-line no-return-assign
+        newNetwork.nodes.forEach((node: Node) => nodeDegreeDict[node._id] = node.degreeCount);
+
+        // eslint-disable-next-line no-plusplus
+        for (let i = 0; i < newNetwork.edges.length; i++) {
+          // Remove edges that don't match degree criteria
+          const edge = newNetwork.edges[i];
+          // eslint-disable-next-line radix
+          if (nodeDegreeDict[edge._from] < parseInt(degreeRange.value[0]) || nodeDegreeDict[edge._from] > parseInt(degreeRange.value[1])) {
```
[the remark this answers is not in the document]
gotdairyya: `degreeRange` is typecast for `number[]` but when I was testing the code and console logging, sometimes it would return a string. I wasn't sure why it was doing that

src/components/Controls.vue
```diff
@@ -78,6 +79,62 @@ export default defineComponent({
...
+        store.commit.setDegreeRange(range);
+      },
+    });
+
+    watch([degreeRange], () => {
+      // Restore network if min and max are restored
+      if (networkOnLoad.value !== null && degreeRange.value[0] === 0 && degreeRange.value[1] === networkOnLoad.value.nodes.length) {
+        store.dispatch.updateNetwork({ network: networkOnLoad.value });
+      } else {
+        const newNetwork = JSON.parse(JSON.stringify(networkOnLoad.value));
```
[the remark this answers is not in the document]
gotdairyya: :thumbs_up:

src/components/Controls.vue
```diff
@@ -78,6 +79,62 @@ export default defineComponent({
...
+        const nodeSet = new Set([]);
+
+        // Create dict of row nodes and max degrees
+        const nodeDegreeDict: {[key: string]: number } = {};
+        // eslint-disable-next-line no-return-assign
+        newNetwork.nodes.forEach((node: Node) => nodeDegreeDict[node._id] = node.degreeCount);
+
+        // eslint-disable-next-line no-plusplus
+        for (let i = 0; i < newNetwork.edges.length; i++) {
```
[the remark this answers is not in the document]
gotdairyya: oh that's much nicer, thank you!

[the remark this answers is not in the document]
gotdairyya: done

src/components/Controls.vue
```diff
@@ -78,6 +79,62 @@ export default defineComponent({
...
+            i--;
+          } else {
+            // Make a record of valid nodes
+            nodeSet.add(edge._from);
+            nodeSet.add(edge._to);
+          }
+        }
+
+        // eslint-disable-next-line no-plusplus
+        for (let i = 0; i < newNetwork.nodes.length; i++) {
```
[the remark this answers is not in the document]
gotdairyya: done

src/components/Controls.vue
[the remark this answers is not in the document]
gotdairyya: Had to use `// eslint-disable-next-line no-undef`, tracking in issue #390

gotdairyya:
@JackWilb I've updated how the degrees are calculated. Could you help with the type setting? And it should be good after that unless you notice anything else

src/components/Controls.vue
```diff
@@ -78,6 +81,53 @@ export default defineComponent({
     });
     const maxConnections = computed(() => store.state.maxConnections);
 
+    // Calculate node degrees
+    // Create dict of row nodes and max degrees
+    const nodeDegreeDict: {[key: string]: {[key: string]: number }} = {};
+
+    const degreeMax = computed(() => {
+      let maxCounter = 0;
+      if (networkOnLoad.value != null) {
```
JackWilb, line 90: This is null on load for me, which causes the slider to initialize to [0,0]

gotdairyya: I think I fixed this

JackWilb: It seems so!

src/lib/utils.ts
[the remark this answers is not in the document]
JackWilb: Derya showed me that this is a bad idea. We've pivoted a bit from this

src/store/index.ts
```diff
@@ -361,6 +363,43 @@ const {
         }
       }
     },
+
+    setNodeDegreeDict(state) {
+      // Create dict of row nodes and max degrees
+      if (state.networkOnLoad != null) {
+        state.networkOnLoad.nodes.forEach((node) => {
```
JackWilb, line 370: Should this be `networkOnLoad`? If we make it `network`, then it might work better with aggregation, etc. We'd just make sure we call it on load and when aggregating.

src/store/index.ts
```diff
@@ -361,6 +363,43 @@ const {
...
+      if (state.networkOnLoad !== null && degreeRange[0] === 0 && degreeRange[1] === state.maxDegree) {
+        store.dispatch.updateNetwork({ network: state.networkOnLoad });
+      } else if (state.networkOnLoad !== null) {
+        // eslint-disable-next-line no-undef
+        const newNetwork = structuredClone(state.networkOnLoad);
+        const nodeSet: Set<string> = new Set([]);
+
+        // Remove edges that don't match degree criteria
+        newNetwork.edges = newNetwork.edges.filter((edge: Edge) => {
+          if (state.nodeDegreeDict[edge._from] >= degreeRange[0] && state.nodeDegreeDict[edge._from] <= degreeRange[1]) {
```
JackWilb, line 389: This gives us all out edges from nodes that are in the degree range (and all their neighbors). Should we also allow all in edges (especially in the case that we don't care about edge directionality)?

JackWilb: Also, should we include their neighbors or only include them if they are also in the degree range?

src/components/FilterOverlay.vue
```diff
@@ -159,8 +162,17 @@ export default defineComponent({
           const aqlNetwork: Network = promise[0];
 
           if (aqlNetwork.nodes.length !== 0) {
-          // Update state with new network
-            store.dispatch.updateNetwork({ network: aqlNetwork });
+            const nodes = defineNeighbors(aqlNetwork.nodes, aqlNetwork.edges as Edge[]);
```
JackWilb, line 165: It feels weird to have the `defineNeighbors` step here. I think this should just be in the store in the update network function. Let me know what you think about that

src/components/FilterOverlay.vue
```diff
@@ -159,8 +162,17 @@ export default defineComponent({
...
+
+            // Build the network object and set it as the network in the store
+            const aqlNetworkElements: Network = {
+              nodes: nodes as Node[],
+              edges: aqlNetwork.edges as Edge[],
+            };
+
+            // Update state with new network
+            store.commit.setNetworkOnLoad(aqlNetworkElements);
+            store.dispatch.updateNetwork({ network: aqlNetworkElements });
```
JackWilb, line 175: If you accept the other change in this file, these will need to be switched and you'll need to pass another network to `setNetworkOnLoad`
```suggestion
            store.dispatch.updateNetwork({ network: aqlNetworkElements });
            store.commit.setNetworkOnLoad(store.state.network);
```

src/components/Controls.vue
```diff
@@ -77,6 +77,12 @@ export default defineComponent({
       return Object.values(nodeColumnTypes).map((colTypes) => Object.entries(colTypes).filter(([_, colType]) => colType === 'category').map(([varName, _]) => varName)).flat();
     });
     const maxConnections = computed(() => store.state.maxConnections);
+    const maxDegree = computed(() => store.state.maxDegree);
+    const degreeRange = ref([0, maxDegree.value]);
+
+    watch([maxDegree], () => {
+      degreeRange.value = [0, maxDegree.value];
+    });
```
JackWilb, line 85: Can this be made into a computed? I'm not sure if this will have issues with reactivity. It could also cause issues with the value in the html below.

```suggestion
    const degreeRange = computed(() => [0, maxDegree.value]);
```

src/components/Controls.vue
```diff
@@ -375,6 +388,36 @@ export default defineComponent({
             />
           </v-list-item>
 
+          <v-list-item>
+            <v-list-item-content> Degree </v-list-item-content>
+            <v-range-slider
+              v-model="degreeRange"
```
JackWilb, line 394: I wonder if this `v-model` is needed if you switch to using a computed.

JackWilb (changes requested):
This code is looking pretty good so far. I have a couple of suggestions in the comments, but the ones about the computed vs a ref with a watcher might not work.

On another note, the node degree filtering doesn't work with an aggregated network. It also doesn't work with quering, since it just copies networkOnLoad, which is not updated. I'm not sure the best way to make all of this work with that 😅 but maybe it will become obvious as you play some more with it... If not, I'm happy to dig in and see what I can see for simplifying it all

src/store/index.ts
gotdairyya: we want it to be `networkOnLoad` because the filtering updates the `network` and that's how the matrix is redrawn. I can check for aggregation and change the network that is used (same for querying). 

If you have another idea for how to redraw the matrix without updating the network that would be great.

src/store/index.ts
gotdairyya: > This gives us all out edges from nodes that are in the degree range (and all their neighbors). Should we also allow all in edges (especially in the case that we don't care about edge directionality)?

Oh, yeah I can do that based on whether it is a directional matrix or not

- [ ] Account for directionality in filtering

gotdairyya: > Also, should we include their neighbors or only include them if they are also in the degree range?

I don't understand this, sorry

src/components/FilterOverlay.vue
gotdairyya: I don't think so! It is another source of creating new network data which I see as equivalent to when smaller networks are loaded

gotdairyya: > This code is looking pretty good so far. I have a couple of suggestions in the comments, but the ones about the computed vs a ref with a watcher might not work.
> 
> On another note, the node degree filtering doesn't work with an aggregated network. It also doesn't work with quering, since it just copies networkOnLoad, which is not updated. I'm not sure the best way to make all of this work with that 😅 but maybe it will become obvious as you play some more with it... If not, I'm happy to dig in and see what I can see for simplifying it all

To Do:

- [ ] have filtering work on aggregated network
- [ ] have filtering work on queried network

src/store/index.ts
JackWilb: > > Also, should we include their neighbors or only include them if they are also in the degree range?
> 
> I don't understand this, sorry

You add both nodes to `nodeSet` when the from node has the right degree. That means the to node will get included. Is that what we want? Or should we just show the connections between all nodes that have the right degree?

That would change the logic to:
1. Filter the nodes to include only those in the degree range.
2. Find all edges that have from and to in the filtered nodes.

I guess I'm just wondering if we're doing this filtering step backwards as it's written

src/components/FilterOverlay.vue
JackWilb: I wasn't very clear here. What I meant was, don't we want to redefine the neighbors each time we update a network? If so, it would make sense to do that in the store function that handles it. However, it could be that we don't want to do it every time, and that there's a good reason to do it here (I'm thinking aggregation)

src/store/index.ts
gotdairyya: So this is what I did at first and then I realized that we need to track the to nodes or else we may end up having an edge without an edge that is connected to and that will throw an error when the matrix is drawn.

JackWilb: The edge should only be included if both the from and to are in the selected subset. That should fix that problem

gotdairyya:
@JackWilb I made quite a few changes so I will list them out here. I didn't make the ref to a computed, I had tried that and had some issues, if you can look at that that would be great.

**Major Changes:**
- Moved the computation of the node degree dict to the utils
- Added logic for directionality in the matrix
- Filtering works for any combination of aggregated network and/or queried network

gotdairyya:
Jack and I decided to go ahead and disable degree filtering when aggregating the network because both of these methods are meant to make the network smaller. It also removes visual complexity.

**Left to do:**
- [ ] Fix color for filtered children cells
- [ ] Reset rangeSlider when aggregating
- [ ] Remove filtered nodes from aggregation function
- [ ] Check if new design works with queried network (or @JackWilb do we disable it here too)

JackWilb:
> Remove filtered nodes from aggregation function

We could probably reset the network before aggregation

> Check if new design works with queried network (or @JackWilb do we disable it here too)

Querying should reset everything, and give you the queried network

JackWilb (dismissed):
Looks good! I have a couple of suggestions for you to fix up before you merge. It should simplify some checks and remove some duplicated code

src/components/Controls.vue
```diff
@@ -185,6 +193,12 @@ export default defineComponent({
     }
 
     function aggregateNetwork(varName: string) {
+      if (filtered.value) {
+        store.commit.setFilteredNetwork(false);
+        if (networkOnLoad.value !== null) { store.dispatch.updateNetwork({ network: networkOnLoad.value }); }
```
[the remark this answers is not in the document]
gotdairyya: Done

src/lib/utils.ts
```diff
@@ -55,3 +55,31 @@ export function formatShortDate(date: Date) {
...
+export function setNodeDegreeDict(networkPreFilter: Network | null, networkOnLoad: Network | null, queried: boolean, directionalEdges: boolean) {
+  // Determine correct network to use
+  let baseNetwork: Network | null = { nodes: [], edges: [] };
+  // Reset node dict
+  const nodeDegreeDict: {[key: string]: number} = {};
+
+  if (networkPreFilter != null || networkOnLoad !== null) {
+    baseNetwork = queried ? structuredClone(networkPreFilter) : structuredClone(networkOnLoad);
+  }
+  if (baseNetwork !== null) {
```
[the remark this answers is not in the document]
gotdairyya: done

src/store/index.ts
```diff
@@ -362,6 +376,88 @@ const {
...
+    },
+
+    setDegreeEntries(state, degreeObject: { maxDegree: number; nodeDegreeDict: {[key: string]: number}}) {
+      state.maxDegree = degreeObject.maxDegree;
+      state.nodeDegreeDict = degreeObject.nodeDegreeDict;
+    },
+
+    setDegreeNetwork(state, degreeRange: number[]) {
+      // Determine correct network to use
+      let baseNetwork: Network | null = { nodes: [], edges: [] };
```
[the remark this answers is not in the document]
gotdairyya: done

src/lib/utils.ts
```diff
@@ -55,3 +55,31 @@ export function formatShortDate(date: Date) {
...
+    baseNetwork.edges.forEach((edge: Edge) => {
+      if (directionalEdges) {
+        // eslint-disable-next-line no-unused-expressions
+        Object.prototype.hasOwnProperty.call(nodeDegreeDict, edge._from) ? nodeDegreeDict[edge._from] += 1 : nodeDegreeDict[edge._from] = 1;
```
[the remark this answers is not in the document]
gotdairyya: Opened in issue #406

JackWilb (approved):

PR #393, opened 2022-05-25
Fix sorting

### LINKED ISSUE #259: Matrix sorting doesn't really work as intended
It sorts rows and columns when using the sort icons on the column headers, that breaks the functionality

### LINKED ISSUE #260: Allow sort order to sort the lineup view

### LINKED ISSUE #294: Sorting an aggregated network and then expanding a node causes the matrix to reset to unsorted state

JackWilb:
Depends on #391 
Closes #259
Closes #260 

Fixes matrix sorting. You can now sort by node, cluster, etc.

TODOs:
- [x] Fix lineup sorting. I need to figure out how to make the lineup order change and how to get changes from lineup back to the store. (#260, might close after)
- [x] Check that sorting twice causes reset
- [x] Check if this fixes #294 (it does not)
- [x] Fix lineup highlighting and selection (needs translation through the `sortOrder` array)
- [x] Make lineup sort not persistent when sorting in matrix (#397)
- [x] Open an issue for lineup grouping matching the matrix (#398)

### DIFF
--- a/src/components/LineUp.vue
+++ b/src/components/LineUp.vue
@@ -35,6 +35,30 @@ export default defineComponent({
       return 330;
     });
 
+    const sortOrder = computed(() => store.state.sortOrder);
+    const lineupOrder = computed(() => {
+      if (lineup.value === null || [...lineup.value.data.getFirstRanking().getOrder()].length === 0) {
+        return [...Array(network.value?.nodes.length).keys()];
+      }
+      return [...lineup.value.data.getFirstRanking().getOrder()];
+    });
+
+    // If store order has changed, update lineup
+    watch(sortOrder, (newSortOrder) => {
+      if (lineup.value !== null) {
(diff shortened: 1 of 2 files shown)


gotdairyya:
Added pointer CSS to make it clear that the sorting icons act as buttons.

gotdairyya (dismissed):
Looks and works well. 

A couple of things I noticed, neither of which are critical: 
- It's not intuitive what the sort should do in cases where there is aggregation
- The sort treats all row labels as strings which gets funky with numbers

JackWilb:
> Looks and works well.
> 
> A couple of things I noticed, neither of which are critical:
> 
>     * It's not intuitive what the sort should do in cases where there is aggregation

This is tracked in #294 

>     * The sort treats all row labels as strings which gets funky with numbers

I'll fix this before merge

Issue #384, opened 2022-03-21
It's not possible to scroll the lineup popover


JackWilb:
Thanks @curtislisle

PR #396, opened 2022-06-03
Lineup fixes

JackWilb:
Depends on #393 

Fixes a ton of small issues that have been affecting the lineup vis

TODOs:
- [ ] See if we can do #384


gotdairyya (changes requested):
Noticed a few bugs:

- [x] When you filter with LineUp the matrix doesn't filter (#400)
image

- [x] The menu for Lineup is cutoff
image

- [x] If a sort order in LineUp is selected, clicking the sort order in the matrix should override it (currently it does not) (#401)
- [x] If the matrix is aggregated, Lineup should preserve the parent-child groupings (currently it does not) (#402)

PR #407, opened 2022-06-29
Fix some aggregated network bugs

### LINKED ISSUE #314: Fix neighbor highlighting on aggregated networks
I think the issue is with the neighbor calculation. For aggregated nodes, children are not neighbors and for children nodes, aggregated nodes are not neighbors. This causes some slight issues in which nodes get highlighted

### LINKED ISSUE #399: Sync aggregation status between matrix and lineup
When they're aggregated, make them both expand and contract together

### LINKED ISSUE #402: If the matrix is aggregated, Lineup should preserve the parent-child groupings
_Originally posted by @gotdairyya in https://github.com/multinet-app/multimatrix/pull/396#pullrequestreview-1013874636_

### LINKED ISSUE #404: Wrong count when the network is aggregated, expanded, and un-aggregated
On load:
image

After issue title:
image

JackWilb:
Closes #314
Closes #399
Closes #402 
Closes #404 

This PR fixes a small issue with the legend count after aggregating, allows aggregating from lineup, and fixes aggregated neighbor highlighting.

The issue with legend counts was an easy fix that just required cloning the network as it was saved on load and read again to aggregate and reset the network.

To allow aggregation from lineup, we watch for that event and trigger a global aggregation, which clears out the lineup and matrix and loads the new network.

To fix the neighbor highlight, I was going to do something complicated with searching through children, etc., but we just need to recompute neighbors when the network expands/retracts. Since the edges are correct, that fixes things up.

There are no real UI changes. The buttons/highlights just work now

TODOs:
- [x] Fix neighbor highlighting issue #314 
- [x] See if nested sorting is possible (will address in a follow up)

### DIFF
--- a/src/components/ControlPanel.vue
+++ b/src/components/ControlPanel.vue
@@ -24,7 +24,14 @@ export default defineComponent({
   setup() {
     // Template objects
     const showMenu = ref(false);
-    const aggregateBy = ref(undefined);
+    const aggregateBy = computed({
+      get() {
+        return store.state.aggregatedBy;
+      },
+      set(value: string | undefined) {
+        store.commit.setAggregatedBy(value);
+      },
+    });
     const directionalEdges = computed({
       get() {
         return store.state.directionalEdges;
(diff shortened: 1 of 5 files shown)


gotdairyya (approved):
Looks great! Also great that networkOnLoad is now a clone of the network. I bet that was having unintended consequences in a lot of places...

PR #409, opened 2022-07-07
Prefer structured clone

### LINKED ISSUE #390: Update eslint
We started using structuredClone and eslint hasn't released the newest version that has the updated globals lib.

JackWilb:
Closes #390

Fixes the node type issues, etc. from structuredClone

N/A

TODOs:
- [x] Figure out the best way to make this work with the open PRs


src/components/IntermediaryNodes.vue
```diff
@@ -52,7 +52,7 @@ export default defineComponent({
       if (network.value !== null && connectivityPaths.value.nodes.length > 0) {
         let sortConnectivityPaths = [];
         const sortKey = intAggregatedBy.value === undefined ? '_key' : intAggregatedBy.value;
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any, vue/no-side-effects-in-computed-properties
         sortConnectivityPaths = connectivityPaths.value.nodes.sort((a: any, b: any) => (a[sortKey] > b[sortKey] ? 1 : -1));
```
gotdairyya, line 56: Does this account for categorical sorting too? Or numbers as strings?

JackWilb: There's no change to the logic, so the answer is however it behaves at the moment is how it will continue to behave.

Numbers will be sorted as strings, e.g. 1, 10, 11, 2, 3, 35, 4, 5. If that's not what we want, then we should open an issue, for sure

PR #410, opened 2022-07-11
Remove cycles from connectivity queries

JackWilb:
No
Found here and potentially before: https://github.com/multinet-app/multimatrix/pull/409#pullrequestreview-1034719675

Removes the cycles from the connectivity query. This should stop the start and end node being the same in multi hops. I have removed all 2 hop and 3 hop cycles. These are the only ones that can be present with our current number of hops (max = 3).

I also fixed an issue with the node/edge attribute selection in the query. Previously we would select `n1.attribute`, but for some variable names that causes problems. I've updated it so we select ```n1.`attribute` ```

Changes are in the query so will change query results, but there are no UI changes.

TODOs:
- [x] Check for 3 hop cycles, 4 hop cycles, etc.
- [x] Test that Crystal's broken query now works
- [x] Allow same start and end nodes
- [x] Why is the legend broken? (weird local dev bug)


gotdairyya:
> Test Query for 2 Hop: This is the go-to query for Graffinity. Works well with these recent changes to Multinet querying. 

In comparison, if the middle node is left to be an approximate match, I get the spinning wheel of doom
image

JackWilb:
Should this be a toggle? There are potentially cases where we'd like the cycle to show. Maybe we even want just the start and end the same and no cycles in the middle?

gotdairyya (approved):
Looks good!

PR #413, opened 2022-07-20
Make matrix and lineup stay in sync when sorting and fix additional sorting related bugs

### LINKED ISSUE #263: Fix lineup selection
Selecting more than one row doesn't work on lineup (and maybe it shouldn't)

When I try to select more than one row it crashes my browser

### LINKED ISSUE #397: Make lineup sort not persistent when sorting in matrix
When lineup is sorted, it overrides the matrix sorts. When the matrix is sorted, remove the lineup sort

### LINKED ISSUE #401: If a sort order in LineUp is selected, clicking the sort order in the matrix should override it
_Originally posted by @gotdairyya in https://github.com/multinet-app/multimatrix/pull/396#pullrequestreview-1013874636_

JackWilb:
Closes #401 
Closes #397
Closes #263 (no more crashing)

This PR addresses a couple of issue that we had with lineup sorting, hovering, and some configuration.

I made the sorting much more robust, while not fully refactoring the code. Lineup now tracks a `permutingMatrix`, which stores lineup's current difference in sort order from the matrix. We need to track this, because when lineup loads with a specific order and is sorted, we need to track the delta. When the matrix is sorted, we can just replace the lineup data so it stays in sync. This approach is much more successful than my previous attempts.

When I added this `permutingMatrix`, it broke the hovering. As a fix, I use the `permutingMatrix` as the map for which nodes are hovered. 

There was also an old function in the lineup component that was supposed to track if multiple rows had been selected. That never worked, because of the way we use selection to hover multiple rows at once (thus taking away our ability to select multiple rows in the lineup). I removed the function for now, since it was conflicting with the hovering logic.

Finally, I grouped all the lineup configuration into one block, and disabled animations, which makes it feel much more snappy.

Same as before, sorting just works now.

TODOs:
- [x] The usual testing

### DIFF
--- a/src/components/LineUp.vue
+++ b/src/components/LineUp.vue
@@ -49,6 +49,7 @@ export default defineComponent({
       return possibleHeight < 500 ? 500 : possibleHeight;
     });
 
+    let lineupIsSorter = false;
     const sortOrder = computed(() => store.state.sortOrder);
     const lineupOrder = computed(() => {
       if (lineup.value === null || [...lineup.value.data.getFirstRanking().getOrder()].length === 0) {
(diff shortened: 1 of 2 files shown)


gotdairyya:
All the filtering and sorting work!

Do you want to control filtering with LineUp?

gotdairyya (approved):

JackWilb:
> All the filtering and sorting work!
> 
> Do you want to control filtering with LineUp?

Yeah, it should work with all the buttons. I think we have open issues for the filtering, though

PR #416, opened 2022-07-23
Degree filter on large query

gotdairyya:
Closes #

At the moment we don't have any way to display results from queries that are > 100.
Adding a check to see how large the query result is will trigger a filtered matrix.

TODOs:
- [ ] Find a semantically meaningful min degree value
- [ ] Change the hover text for filter to show # of children

### DIFF
--- a/src/components/ConnectivityQuery.vue
+++ b/src/components/ConnectivityQuery.vue
@@ -569,6 +569,10 @@ export default defineComponent({
             store.commit.setDirectionalEdges(true);
             store.commit.setQueriedNetworkState(true);
             store.commit.setDegreeEntries(setNodeDegreeDict(store.state.networkPreFilter, store.state.networkOnLoad, store.state.queriedNetwork, store.state.directionalEdges));
+            if (promise.length >= 100) {
+              store.commit.setMinDegree(5);
+              store.commit.setDegreeNetwork([5, store.state.maxDegree]);
+            }
           } else {
             // Update state with empty network
             store.dispatch.aggregateNetwork(undefined);
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -78,6 +78,7 @@ const {
     selectedHops: 1,
     nodeDegreeDict: {},
     maxDegree: 0,
+    minDegree: 0,
     networkPreFilter: null,
     queriedNetwork: false,
     filteredNetwork: false,
@@ -387,6 +388,10 @@ const {
     setDegreeEntries(state, degreeObject: { maxDegree: number; nodeDegreeDict: {[key: string]: number}}) {
       state.maxDegree = degreeObject.maxDegree;
       state.nodeDegreeDict = degreeObject.nodeDegreeDict;
+    },
+
+    setMinDegree(state, minDegree: number) {
+      state.minDegree = minDegree;
     },
 
     setDegreeNetwork(state, degreeRange: number[]) {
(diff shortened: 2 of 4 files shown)


gotdairyya:
Netlify does not like how I am treating the hover object -- I tried to add a check for it but it didn't work. Line 167, MultiMatrix.vue

gotdairyya:
Test Query: RPC1, 
`Node 0 Label ~= CBB`

src/store/index.ts
```diff
@@ -389,6 +390,10 @@ const {
       state.nodeDegreeDict = degreeObject.nodeDegreeDict;
     },
 
+    setMinDegree(state, minDegree: number) {
```
JackWilb, line 393: We don't have a setMaxDegree, why can't we just get this info from `setDegreeNetwork` or another `setDegreeEntries`?

src/components/ConnectivityQuery.vue
```diff
@@ -569,6 +569,24 @@ export default defineComponent({
...
+
+              // Set min value if the node degree occurence < 100
+              Object.entries(degreeCount).every(([degree, occurence]) => {
+                if (occurence < 100) {
+                  store.commit.setMinDegree(Number(degree));
+                  store.commit.setDegreeNetwork([Number(degree), store.state.maxDegree]);
+                  return false;
+                } return true;
+              });
+            }
```
JackWilb, line 589: Does doing this after setting the network object cause it to render twice? Maybe we want to find the degrees that work first and then update the network object?

src/components/ConnectivityQuery.vue
```diff
@@ -569,6 +569,24 @@ export default defineComponent({
...
+              }
+
+              // Set min value if the node degree occurence < 100
+              Object.entries(degreeCount).every(([degree, occurence]) => {
+                if (occurence < 100) {
+                  store.commit.setMinDegree(Number(degree));
+                  store.commit.setDegreeNetwork([Number(degree), store.state.maxDegree]);
+                  return false;
+                } return true;
+              });
```
JackWilb, line 588: This should help simplify a bit. We're not using the boolean results from the `every`, so I think `forEach` is the way to go.

```suggestion
              Object.entries(degreeCount).forEach(([degree, occurrence]) => {
                if (occurrence < 100) {
                  store.commit.setMinDegree(Number(degree));
                  store.commit.setDegreeNetwork([Number(degree), store.state.maxDegree]);
                }
              });
```

JackWilb (commented):
There are some questions that I have before an approval. Thanks for this feature

src/components/ConnectivityQuery.vue
gotdairyya: Will your suggestion keep going through all the entries? I'm using `every` so that it stops the `for loop` once the conditional is met

JackWilb: Oh, that makes sense. I think your approach is better. Would you just leave a comment above the `every` call, to remind me when I see this again? :)

gotdairyya: yup!

src/store/index.ts
gotdairyya: I want to keep it separate because in `ControlPanel.vue` minDegree is a computed value and it should be 0 unless it is a large query

src/components/ConnectivityQuery.vue
gotdairyya: great suggestion

gotdairyya:
Created issue #417 because although this solves larger query results, it is a not a solution for very large query returns that crash the browser.

Example query:
workspace: `marclab`
network: `RC1---`
query: `node0.Label ~= CBB`

src/store/index.ts
JackWilb: We met to discuss this. There are likely some edge cases that this will break

PR #420, opened 2022-08-05
Bump vue to 2.7 and ts to 4.7

JackWilb:
No

This PR raises the version of vue and typescript to get us on to the most modern versions, 2.7 and 4.7. This will make out migration to vue 3 much easier once it its available. The code is also simplified here using `script setup`. Now, there's no need to return the variables for the template to use at the end of the script block, all variables are just available to the template.

N/A

TODOs:
- [x] Test that nothing has broken


JackWilb:
I need to push a fix to `nodeDegreeDict` on monday

gotdairyya (approved):
Played around and everything looks good!

